PR #60, opened 2019-07-02
added report api url in Pay API

saravankumarpa:
- renamed receipt to report api in oc scripts

- added report api url in payapi dc


pwei1018:
Review done.

PR #188, opened 2020-04-16
Changes for moving error codes to table

sumesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/2890

*Description of changes:*
- Created a table to store all error messages
- Created 2 endpoints:
```
GET /codes/errors
GET /codes/errors/<code>

```
- Added caching to load the errors on app startup
- Changed all errors to use data from table in error message
- Changed all error to match with schema 'problem'
- Changed error content type for all to application/problem+json

*Checklist:*
- [ ] No lint errors


pay-api/src/pay_api/exceptions/__init__.py
```diff
@@ -45,7 +46,8 @@ def error_to_response(error: Error, invalid_params=None):
 
 def construct_type(code):
     """Construct type in response."""
-    return 'https://bcrs.gov.bc.ca/.well_known/schemas/problem#{}'.format(code)
+    # return 'https://bcrs.gov.bc.ca/.well_known/schemas/problem#{}'.format(code)
```
sutherlanda, line 49: Stray comment

sumesh-aot: Kept it there, as we may need to re-consider when we move all error codes (all API) to databse.

sutherlanda: Got it 👍

pay-api/migrations/versions/6ba89cd39cc8_adding_error_codes_table.py
```diff
@@ -0,0 +1,137 @@
...
+revision = '6ba89cd39cc8'
+down_revision = '8eac44955fb6'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    error_code_table = op.create_table('error_code',
+                                       sa.Column('code', sa.String(length=50), nullable=False),
```
saravanpa-aot, line 21: the only minor thing i can think is to have some kind of system identifier ;either as a column or as prefix .something  PAY_INVALID_IDENITIFER.
When we move to the common system for errors , there will be huge number of errors

saravanpa-aot, line 21: the only minor thing i can think is to have some kind of system identifier ;either as a column or as prefix .something  PAY_INVALID_IDENITIFER.
When we move to the common system for errors , there will be huge number of errors

pay-api/src/pay_api/__init__.py
```diff
@@ -90,3 +91,13 @@ def shell_context():
         return {'app': app, 'jwt': jwt, 'db': db, 'models': models}  # pragma: no cover
 
     app.shell_context_processor(shell_context)
+
+
+def build_cache(app):
```
saravanpa-aot, line 96: may be make it as private?

pay-api/tests/unit/api/test_transaction.py
```diff
@@ -61,7 +61,7 @@ def test_transaction_post_with_invalid_return_url(session, client, jwt, app):
     }
     rv = client.post(f'/api/v1/payment-requests/{payment_id}/transactions', data=json.dumps(data),
                      headers={'content-type': 'application/json'})
-    assert rv.status_code == 401
+    assert rv.status_code == 400
```
saravanpa-aot, line 64: this change is fine rite? 401 to 400?

pay-api/migrations/versions/6ba89cd39cc8_adding_error_codes_table.py
sumesh-aot: Hmm.. I think when we move to common table, there would be an indicator for the system (pay, auth, legal etc.). Do we need to incorporate this now, as this is in our schema ?

pay-api/src/pay_api/__init__.py
sumesh-aot: Yes, will change in next PR if that's ok.

pay-api/migrations/versions/6ba89cd39cc8_adding_error_codes_table.py
saravanpa-aot: yeah.. we might not need now.

PR #241, opened 2020-06-15
Don't display temproary incorpration number in receipt.

pwei1018:
*Issue #:*
https://github.com/bcgov/entity/issues/3968

*Description of changes:*
Don't include temproary incorpration number in invoice dict object.
Display service fee in receipt.

*Checklist:*
- [ ] No lint errors


report-api/report-templates/payment_receipt.html
```diff
@@ -432,10 +450,10 @@
         <td colspan="2" class="receipt-footer">
           <div class="footer-info">
             {% if filingIdentifier %}
-              <span>Filing #{{filingIdentifier}}</span>
+            <span>Filing #{{filingIdentifier}}</span>
             {% endif %}
             {% if invoice['businessIdentifier'] %}
-              <span>Incorporation #{{invoice['businessIdentifier']}}</span>
+            <span>Incorporation #{{invoice['businessIdentifier']}}</span>
```
sumesh-aot, line 456: Since we are popping business identifier from the response, probably we will need a None safe get here ?

pwei1018: It seems the template if statement already handle the None type. Do you mean we should handle "businessIdentifier": "None" in payload?

sumesh-aot: What I meant was;
`if invoice['businessIdentifier']` will throw error if the key `businessIdentifier` is not present in invoice dict.
Changing it to;
`if invoice.get('businessIdentifier', None)` will make it as None safe key.

pwei1018: Fixed.

PR #286, opened 2020-08-14
Update invoice status on changes to distribution code

sumesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/4373

*Description of changes:*
Update invoice status on changes to distribution code

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/pay-api/src/pay_api/models/invoice.py
+++ b/pay-api/src/pay_api/models/invoice.py
@@ -17,12 +17,12 @@ from marshmallow import fields, post_dump
 from sqlalchemy import ForeignKey
 from sqlalchemy.orm import relationship
 
-from pay_api.utils.enums import LineItemStatus
+from pay_api.utils.enums import InvoiceStatus, LineItemStatus
 from .audit import Audit
 from .base_schema import BaseSchema
 from .db import db, ma
 from .invoice_reference import InvoiceReferenceSchema
-from .payment_line_item import PaymentLineItemSchema
+from .payment_line_item import PaymentLineItem, PaymentLineItemSchema
 from .receipt import ReceiptSchema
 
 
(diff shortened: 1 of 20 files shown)


pay-api/src/pay_api/services/payment_transaction.py
```diff
@@ -310,7 +310,7 @@ def update_transaction(payment_identifier: int, transaction_id: uuid,  # pylint:
         except ServiceUnavailableException as exc:
             txn_reason_code = exc.status
             receipt_details = None
-
+        print('0000000', receipt_details)
```
pwei1018, line 313: need print?

pay-api/src/pay_api/models/invoice.py
```diff
@@ -76,6 +76,20 @@ def find_by_id_and_payment_id(cls, identifier: int, pay_id: int):
...
+    @classmethod
+    def update_invoices_for_revenue_updates(cls, fee_distribution_id: int):
+        """Find and update all invoices using the distribution id."""
+        query = db.session.query(Invoice) \
+            .join(PaymentLineItem, PaymentLineItem.invoice_id == Invoice.id) \
+            .filter(PaymentLineItem.fee_distribution_id == fee_distribution_id)
+
+        invoices: [Invoice] = query.all()
+        for invoice in invoices:
+            if invoice.invoice_status_code == InvoiceStatus.PAID.value:
```
saravanpa-aot, line 88: shouldn't we do for Unpaid invoices as well? not sure if thats gonna happen for direct pay..but was thinking of user going to bambora screen and closing screen kind of edge cases... 
also should we have a date check as well? like may be 1 week or 1 month time ; will records before that have to be rectified in paybc?

pay-api/src/pay_api/services/payment_transaction.py
sumesh-aot: Will remove this in next PR

pay-api/src/pay_api/models/invoice.py
sumesh-aot: Unpaid invoices we can think about it. Didn't want to handle bunch of 404 from PayBC. The job to update stale payment will be 30 minutes, so we shouldn't see lot of those issues. 
 
The job will check the GL status before updating the revenue account. This feature, we are considering only for any error in GL code. If the use case is for new GL code, we should be creating new GL code and map the filing types to GL code.

saravanpa-aot: yup.. thats good ..

PR #344, opened 2020-09-22
Updating pay-api spec with draft changes for API gateway publishing

sumesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/4942

*Description of changes:*
- Updating pay-api spec with draft changes for API gateway publishing
- Only first draft, subject to review and revisions
- Fixed lint errors
- Modified the nginx conf to support new example naming 

*Checklist:*
- [ ] No lint errors


doug-lovett (commented):
Should a request with a path parameter include a 404 not found error response?
Should a POST/PATCH/PUT request with a body include a 400/422 invalid payload error response?  
For the /api/v1/accounts/{accountId}/payments/queries request examples the date format is different than PPR: PPR uses the ISO format "2015-03-25". It looks like pay also uses the ISO format in the responses.
For responses where the format is not JSON, it would be nice to include examples using externalValue. Unfortunately this is not working with my version of StopLight (2.0.0). I am getting a "value property does not exist" error with:
              examples:
                test-external-value:
                  summary: 'Test if externalValue example works'
                  externalValue: 'https://fb-apigw-demo.web.app/ppr-api-property-lengths.pdf'

For GET /api/v1/payment-requests/{payment_identifier}/receipts the Content-Type is application/pdf but the examples are JSON.
Adding a summary to an example shows up as a description in the Apigee generated version. 
If the apiKey is a gateway only security requirement then I think it can be removed from your version of the spec.

sumesh-aot:
@doug-lovett I have incorporated some review comments. Adding PDF as an example is causing prism to be exited without much explanation. I will investigate later.

doug-lovett (approved):
Just one comment: change "statment" to "statement" for the path:
/api/v1/accounts/{accountId}/statments/{statementId}
The gateway published version is available from https://fb-apigw-demo.web.app/pay-api or directly from https://bcregistry-demo.apigee.io/docs/payment-proxy/1/overview
The gateway sandbox env pay proxy is pointing at the mock service which uses the spec example responses.

sumesh-aot:
> Just one comment: change "statment" to "statement" for the path:
> /api/v1/accounts/{accountId}/statments/{statementId}

Oops, I will fix it and update this PR. :)

PR #433, opened 2020-11-25
5514 add account number masking to pay-api

sutherlanda:
*Issue #:* 5514
https://github.com/bcgov/entity/issues/5514

*Description of changes:*

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/pay-api/src/pay_api/services/payment_account.py
+++ b/pay-api/src/pay_api/services/payment_account.py
@@ -29,7 +29,7 @@ from pay_api.utils.enums import PaymentSystem, StatementFrequency, PaymentMethod
 from pay_api.utils.errors import Error
 from pay_api.utils.user_context import user_context, UserContext
 from pay_api.utils.util import get_str_by_path, current_local_time, \
-    get_local_formatted_date
+    get_local_formatted_date, mask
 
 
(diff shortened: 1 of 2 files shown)


sumesh-aot (commented):
There are lint errors, and missing test cases. As discussed, we need to handle this based on the user role.

sutherlanda:
> There are lint errors, and missing test cases. As discussed, we need to handle this based on the user role.

The listing error (singular) is corrected. 

The last discussion we had on this, you were going to check with Saravan:

"I think currently the auth system calls this service with system account, that also needs to be changed to pass in user token
need to check with saravan"

If you would like me to do that, just say so, but as of now I am no aware of what solution you would like to be put in place.  I will add test cases once there is a full solution.

sumesh-aot:
> "I think currently the auth system calls this service with system account, that also needs to be changed to pass in user token

Yes, that's correct. We need to factor in changes to permissions in auth side. Also auth needs to pass in the user token instead of a service account token. We don't have a separate ticket to handle this on auth side. 
And as I discussed, this task needs some discussion with team and proper solution starting from the auth side and I don't think the nature of this ticket allows to work independently.

saravanpa-aot:
This task is handled in other tickets.so closing the PR for now.

PR #470, opened 2020-12-10
Role based masking added

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/5563

*Description of changes:*

Follow up PR is https://github.com/bcgov/sbc-auth/pull/1199

1) called the auth authorisation endpoint.
2) added the permissions to user context so that it easy to look up
3) added mask function
4) added mask to the events

### DIFF
--- a/jobs/payment-jobs/tasks/cfs_create_account_task.py
+++ b/jobs/payment-jobs/tasks/cfs_create_account_task.py
@@ -24,6 +24,7 @@ from pay_api.services.oauth_service import OAuthService
 from pay_api.services.queue_publisher import publish_response
 from pay_api.utils.constants import RECEIPT_METHOD_PAD_DAILY
 from pay_api.utils.enums import AuthHeaderType, CfsAccountStatus, ContentType
+from pay_api.utils.util import mask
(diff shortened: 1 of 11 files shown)


jobs/payment-jobs/tasks/cfs_create_account_task.py
```diff
@@ -119,6 +120,8 @@ def create_accounts(cls):  # pylint: disable=too-many-locals
             pending_account.save()
 
             # Publish message to the Queue, saying account has been created. Using the event spec.
+            payment_info.update(
```
sumesh-aot, line 123: Not part of this ticket, but we will need to move this event publishing to pay-api. Job will be executed only when CFS is up, but we need to send the PAD agreements when the account is created, else the 3 day period won't work.

sumesh-aot (approved):

jobs/payment-jobs/tasks/cfs_create_account_task.py
saravanpa-aot: sure.will handle in the next PR

PR #486, opened 2020-12-17
Fix for adding corp type to support names authorised requests

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/5826

*Description of changes:*

1) just added the corp type as a header.

I fixed that breaking tests.not related to this though.

### DIFF
--- a/pay-api/src/pay_api/services/auth.py
+++ b/pay-api/src/pay_api/services/auth.py
@@ -64,7 +64,7 @@ def check_auth(business_identifier: str, account_id: str = None, corp_type_code:
                                                                      f'/authorizations?expanded=true'
             additional_headers = None
             if corp_type_code:
-                additional_headers = {'corp-type': corp_type_code}
+                additional_headers = {'product-type': corp_type_code}
             auth_response = RestService.get(auth_url, bearer_token, AuthHeaderType.BEARER, ContentType.JSON,
                                             additional_headers=additional_headers).json()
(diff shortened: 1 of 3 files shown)


pay-api/src/pay_api/services/auth.py
```diff
@@ -60,21 +60,17 @@ def check_auth(business_identifier: str, account_id: str = None, corp_type_code:
...
-                roles: list = auth_response.get('roles', [])
-                if roles:
-                    is_authorized = True
-            else:  # For activities not specific to a product
-                auth_url = current_app.config.get('AUTH_API_ENDPOINT') + f'orgs/{account_id}' \
-                                                                         f'/authorizations?expanded=true'
-
-                auth_response = RestService.get(auth_url, bearer_token, AuthHeaderType.BEARER, ContentType.JSON).json()
-                # TODO Create a similar response as in auth response
+                additional_headers = {'corp-type': corp_type_code}
```
sumesh-aot, line 67: may be make the header as product-type or business-type please. Corp Type is actually something we need to refactor on pay side.

sumesh-aot (approved):
Please check the comment

pay-api/src/pay_api/services/auth.py
saravanpa-aot: ok..wil go for product-type

PR #535, opened 2021-01-07
returning message as a list incase of error scenarios

saravanpa-aot:
*Description of changes:*

A fix related to a UI bug. Not necessarily need to be fixed here ;but for the sake of consistency , fixing it here.
service returns the validation messages as a list when CFS is available and validation is carried out successfully.But when the CFS is down , it was returning single string message. Thats been changed. 
The error message when CFS is not supposed to be used by UI ;but because of a bug it's been read and it failed  because it expected a list.


sumesh-aot (approved):
Flake8 errors are fixed in my PR

PR #572, opened 2021-02-08
New job for calling CGI added (#7)

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/6246

*Description of changes:*
Added new job for FEEDBACK and ACK jobs

Test cases are missing as of now. The challenge is since the files are getting moved around , it changes files in the local file system as well ;which makes it a bad test. I might create a tech debt ticket to address this incase if i cant do it in my next PR


jobs/ftp-poller/tasks/cas_poller_ftp.py
```diff
@@ -0,0 +1,79 @@
+# Copyright © 2019 Province of British Columbia
```
saravanpa-aot, line 1: this file got renamed.contents are same

jobs/ftp-poller/invoke_jobs.py
```diff
@@ -64,16 +64,16 @@ def shell_context():
...
     application = create_app()
 
     application.app_context().push()
     if job_name == 'FTP_POLLER':
-        PollFtpTask.poll_ftp()
+        CASPollerFtpTask.poll_ftp()
         application.logger.info(f'<<<< Completed Polling FTP >>>>')
     else:
-        PollFtpTask.poll_ftp()
+        CASPollerFtpTask.poll_ftp()
```
sumesh-aot, line 76: Is this correct, or should it be cgi poller ftp ?

sumesh-aot (approved):
Check my comments

jobs/ftp-poller/invoke_jobs.py
saravanpa-aot: Corrected it.Thanks for catching it

PR #703, opened 2021-05-26
Added disbursement delay

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/7061

*Description of changes:*

Added disbursement delay while fetching invoices.
Invoices are connected to receipts and avoided the invoices which has a receipt after disbursement date [thats why the negation operator]

### DIFF
--- a/jobs/payment-jobs/tasks/ejv_partner_distribution_task.py
+++ b/jobs/payment-jobs/tasks/ejv_partner_distribution_task.py
@@ -14,7 +14,7 @@
 """Task to create Journal Voucher."""
 
 import time
-from datetime import datetime
+from datetime import datetime, timedelta
 from typing import List
 
 from flask import current_app
@@ -27,8 +27,10 @@ from pay_api.models import EjvInvoiceLink as EjvInvoiceLinkModel
 from pay_api.models import FeeSchedule as FeeScheduleModel
 from pay_api.models import Invoice as InvoiceModel
 from pay_api.models import PaymentLineItem as PaymentLineItemModel
+from pay_api.models import Receipt as ReceiptModel
 from pay_api.models import db
 from pay_api.utils.enums import DisbursementStatus, InvoiceStatus, PaymentMethod
+from sqlalchemy import Date, cast
 
 from tasks.common.cgi_ejv import CgiEjv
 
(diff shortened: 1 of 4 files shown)


jobs/payment-jobs/tasks/ejv_partner_distribution_task.py
```diff
@@ -189,12 +191,14 @@ def _find_line_items_by_invoice_and_distribution(cls, distribution_code_id, invo
...
     def _get_invoices_for_disbursement(cls, partner):
         """Return invoices for disbursement."""
+        disbursement_date = datetime.today() - timedelta(days=current_app.config.get('DISBURSEMENT_DELAY_IN_DAYS'))
         invoices: List[InvoiceModel] = db.session.query(InvoiceModel) \
             .filter(InvoiceModel.invoice_status_code == InvoiceStatus.PAID.value) \
             .filter(
             InvoiceModel.payment_method_code.notin_([PaymentMethod.INTERNAL.value, PaymentMethod.DRAWDOWN.value])) \
             .filter((InvoiceModel.disbursement_status_code.is_(None)) |
                     (InvoiceModel.disbursement_status_code == DisbursementStatus.ERRORED.value)) \
+            .filter(~InvoiceModel.receipts.any(cast(ReceiptModel.receipt_date, Date) >= disbursement_date.date())) \
```
sumesh-aot, line 201: Shouldn't this be **<=** ? 
For e.g, if a PAD payment receipt is created for Monday, then when the job runs on Friday it should pick all invoices which are PAID and have receipt date less than or equal to monday right ? 🤔

saravanpa-aot: hm..I applied the ~ , the negation..without negation , we can do the logic <=.. so the logic checks now ;if there is no receipt after the disbursement date ...
The only reasoning for that is incase if there are multiple receipts and one of them is after disbursement date , it shudnt disbursed...

saravanpa-aot: <= is good enough if no multiple receipts scenario is in place..

sumesh-aot: didn't notice the **~**.

sumesh-aot (approved):

PR #730, opened 2021-07-14
Adding PATCH endpoint for routing slips

sumesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/7795

*Description of changes:*
- PATCH endpoint for routing slip to update the status

### DIFF
--- a/pay-api/src/pay_api/services/fas/routing_slip.py
+++ b/pay-api/src/pay_api/services/fas/routing_slip.py
@@ -24,7 +24,7 @@ from pay_api.models import PaymentAccount as PaymentAccountModel
 from pay_api.models import RoutingSlip as RoutingSlipModel
 from pay_api.models import RoutingSlipSchema
 from pay_api.services.cfs_service import CFSService
-from pay_api.utils.enums import CfsAccountStatus, PaymentStatus, PaymentSystem, RoutingSlipStatus
+from pay_api.utils.enums import CfsAccountStatus, PatchActions, PaymentStatus, PaymentSystem, RoutingSlipStatus
 from pay_api.utils.errors import Error
 from pay_api.utils.user_context import user_context
 from pay_api.utils.util import string_to_date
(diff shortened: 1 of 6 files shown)


karthik-aot (approved):

pay-api/src/pay_api/services/fas/routing_slip.py
```diff
@@ -233,3 +233,19 @@ def create(cls, request_json: Dict[str, any], **kwargs):
...
+    def update(cls, rs_number: str, action: str, request_json: Dict[str, any]) -> Dict[str, any]:
+        """Update routing slip."""
+        if (patch_action := PatchActions.from_value(action)) is None:
+            raise BusinessException(Error.PATCH_INVALID_ACTION)
+
+        routing_slip: RoutingSlipModel = RoutingSlipModel.find_by_number(rs_number)
+        if routing_slip is None:
+            raise BusinessException(Error.FAS_INVALID_ROUTING_SLIP_NUMBER)
+
+        if patch_action == PatchActions.UPDATE_STATUS:
```
saravanpa-aot, line 247: hm... so how will the updates work?
will we have more patchactions?

saravanpa-aot: i was thinking if we should just update based on the field?

saravanpa-aot (approved):

pay-api/src/pay_api/services/fas/routing_slip.py
sumesh-aot: That was added to avoid the complexity we are seeing in auth-api right now. 
So that a block of value can be sent, and based on action selected values gets updated. 
https://restful-api-design.readthedocs.io/en/latest/methods.html#patch-vs-put the standards also suggest to go with a formal approach. 
And if we had similar thing in auth-api things would have been much simpler I think.
And we have only update status atm, but things can change quickly as we saw in all of our apps.

PR #749, opened 2021-08-13
misc tests including folio search and a model test 

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/7796

*Description of changes:*

1) added api test with folio search
2) added a base model test 
3) added faker so that we don't have to supply arbitrary values

### DIFF
--- a/pay-api/src/pay_api/models/payment_account.py
+++ b/pay-api/src/pay_api/models/payment_account.py
@@ -63,7 +63,7 @@ class PaymentAccountSchema(BaseSchema):  # pylint: disable=too-many-ancestors
         """Returns all the fields from the SQLAlchemy class."""
 
         model = PaymentAccount
-        exclude = ['versions', 'pad_activation_date', 'id']
+        exclude = ['versions', 'pad_activation_date']
 
     payment_method = fields.String(data_key='payment_method')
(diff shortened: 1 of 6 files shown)


pay-api/src/pay_api/models/payment_account.py
```diff
@@ -63,7 +63,7 @@ class Meta:  # pylint: disable=too-few-public-methods
         """Returns all the fields from the SQLAlchemy class."""
 
         model = PaymentAccount
-        exclude = ['versions', 'pad_activation_date', 'id']
+        exclude = ['versions', 'pad_activation_date']
```
saravanpa-aot, line 66: one of the tests needs the payment account id in the response ; 
Lmk if its okay to expose the id , If not i can find other ways to get the id...

sumesh-aot: ok, id may confuse if it's included in the response as there would be both id and accountId in paymentAccount resource.

sumesh-aot (approved):

pay-api/src/pay_api/models/payment_account.py
saravanpa-aot: will fix it in the next PR

PR #762, opened 2021-08-31
added corp type and filing type to fee schedules endpoint

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/8631

*Description of changes:*

- Added corp type and filing type details into schema
- Added a description search.. the search is not perfect ;I will improve it later.Put a TODO for this.. 

### DIFF
--- a/pay-api/src/pay_api/models/fee_schedule.py
+++ b/pay-api/src/pay_api/models/fee_schedule.py
@@ -93,14 +93,14 @@ class FeeSchedule(db.Model):
             query = query.filter_by(corp_type_code=corp_type_code)
 
         if description:
-            # description is free text search
+            # TODO arrive at a better search
             description_list = description.replace(' ', '%')
             query = query.join(CorpType,
                                CorpType.code == FeeSchedule.corp_type_code). \
                 join(FilingType, FilingType.code == FeeSchedule.filing_type_code)
             query = query.filter(
-                or_(func.lower(FilingType.description).contains(description_list.lower(), autoescape=True),
-                    func.lower(CorpType.description).contains(description_list.lower(), autoescape=True)))
+                or_(func.lower(FilingType.description).contains(description_list.lower()),
+                    func.lower(CorpType.description).contains(description_list.lower())))
 
         return query.all()
 
--- a/pay-api/src/pay_api/models/fee_schedule.py
+++ b/pay-api/src/pay_api/models/fee_schedule.py
@@ -14,14 +14,15 @@
 """Model to handle all operations related to fee and fee schedule."""
 
 from datetime import date, datetime
+from operator import or_
 
-from sqlalchemy import ForeignKey
+from sqlalchemy import ForeignKey, func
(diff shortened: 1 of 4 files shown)


pay-api/src/pay_api/models/fee_schedule.py
```diff
@@ -91,6 +92,16 @@ def find_all(cls, corp_type_code: str = None, filing_type_code: str = None):
...
             query = query.filter_by(corp_type_code=corp_type_code)
 
+        if description:
+            # description is free text search
+            description_list = description.replace(' ', '%')
+            query = query.join(CorpType,
+                               CorpType.code == FeeSchedule.corp_type_code). \
+                join(FilingType, FilingType.code == FeeSchedule.filing_type_code)
+            query = query.filter(
+                or_(func.lower(FilingType.description).contains(description_list.lower(), autoescape=True),
```
sumesh-aot, line 102: Shouldn't this be using `like` rather than `contains` ? with autoescape can `contains` handle multiple `%` ?

pay-api/src/pay_api/models/fee_schedule.py
```diff
@@ -91,6 +92,16 @@ def find_all(cls, corp_type_code: str = None, filing_type_code: str = None):
         if corp_type_code:
             query = query.filter_by(corp_type_code=corp_type_code)
 
+        if description:
+            # description is free text search
+            description_list = description.replace(' ', '%')
```
sumesh-aot, line 97: may be have a better name than `_list`, I thought it's a list and was wondering how `lower()' function would work on list. Then realized it's not list but a string.

pay-api/src/pay_api/resources/fee_schedule.py
```diff
@@ -19,12 +19,9 @@
 
 from pay_api.exceptions import BusinessException
```
sumesh-aot, line 20: Missing test cases.

sumesh-aot (approved):
Good to merge once test cases are added. Also please see the comments.

karthik-aot (approved):

pay-api/src/pay_api/models/fee_schedule.py
saravanpa-aot: hm..yeah..i am planning to rewrite this...  i was trying to split the string and do a like in both of the fields ; i think thats the way to go...the challenge with multiple % is basically we are searching same string in two columns ...might not be present in one column ;but in other... so have to figure it out...   I will definitely rewrite it after the core functionality of payment is done..

pay-api/src/pay_api/models/fee_schedule.py
saravanpa-aot: changed

pay-api/src/pay_api/resources/fee_schedule.py
saravanpa-aot: added test case

PR #764, opened 2021-09-07
auth_response None

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/8659

*Description of changes:*

When the businessId is empty , we wont call auth and auth_response will be None.so auth_response is set to empty dict to avoid breaking further code.

### DIFF
--- a/pay-api/tests/unit/api/test_payment_request.py
+++ b/pay-api/tests/unit/api/test_payment_request.py
@@ -31,8 +31,9 @@ from pay_api.utils.enums import InvoiceStatus, PaymentMethod, Role
 from tests.utilities.base_test import (
     activate_pad_account, get_basic_account_payload, get_claims, get_gov_account_payload, get_payment_request,
     get_payment_request_for_wills, get_payment_request_with_folio_number, get_payment_request_with_no_contact_info,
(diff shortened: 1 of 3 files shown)


pay-api/tests/unit/api/test_payment_request.py
```diff
@@ -316,9 +317,13 @@ def test_payment_creation_with_routing_slip(session, client, jwt, app):
...
 
-def test_payment_creation_with_existing_routing_slip(client, jwt):
+@pytest.mark.parametrize('payment_requests', [
+    get_payment_request(),
+    get_payment_request_without_bn()
+])
+def test_payment_creation_with_existing_routing_slip(client, jwt, payment_requests):
     """Assert that the endpoint returns 201."""
-    claims = get_claims(roles=[Role.FAS_CREATE.value, Role.FAS_SEARCH.value, Role.STAFF.value])
+    claims = get_claims(roles=[Role.FAS_CREATE.value, Role.FAS_SEARCH.value, Role.STAFF.value, 'make_payment'])
```
saravanpa-aot, line 326: Had to give make_payment to this test user since in this case auth is not invoked to get the staff roles.Change only in test.

sumesh-aot (approved):

PR #814, opened 2021-11-10
party name changed to account id instead of account name

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/9561

*Description of changes:*

- changes to party name being account id

### DIFF
--- a/pay-api/src/pay_api/config.py
+++ b/pay-api/src/pay_api/config.py
@@ -105,7 +105,7 @@ class _Config():  # pylint: disable=too-few-public-methods
     CFS_ACCOUNT_DESCRIPTION = _get_config('CFS_ACCOUNT_DESCRIPTION', default='BCR')
     CFS_INVOICE_PREFIX = os.getenv('CFS_INVOICE_PREFIX', 'REG')
     CFS_RECEIPT_PREFIX = os.getenv('CFS_RECEIPT_PREFIX', 'RCPT')
-    CFS_PARTY_PREFIX = os.getenv('CFS_RECEIPT_PREFIX', '')
+    CFS_PARTY_PREFIX = os.getenv('CFS_PARTY_PREFIX', 'BCR-')
 
     # PAYBC Direct Pay Settings
     PAYBC_DIRECT_PAY_REF_NUMBER = _get_config('PAYBC_DIRECT_PAY_REF_NUMBER')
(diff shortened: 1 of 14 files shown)


pay-api/src/pay_api/services/cfs_service.py
```diff
@@ -38,21 +38,23 @@ class CFSService(OAuthService):
...
     @classmethod
-    def create_cfs_account(cls, name: str, contact_info: Dict[str, Any],
-                           payment_info: Dict[str, any] = None, receipt_method: str = None) -> Dict[str, str]:
+    def create_cfs_account(cls, identifier: str, contact_info: Dict[str, Any],  # pylint: disable=too-many-arguments
+                           payment_info: Dict[str, any] = None,
+                           receipt_method: str = None, site_name=None) -> Dict[str, str]:
         """Create a cfs account and return the details."""
-        name = re.sub(r'[^a-zA-Z0-9]+', ' ', name)
+        party_prefix = current_app.config.get('CFS_PARTY_PREFIX')
+        party_id = re.sub(r'[^a-zA-Z0-9]+', ' ', f'{party_prefix}{identifier}')  # TODO is the prefixing needed?
```
saravanpa-aot, line 46: I am not sure if this is really needed...do we need it?

pay-api/src/pay_api/config.py
```diff
@@ -105,6 +105,7 @@ class _Config():  # pylint: disable=too-few-public-methods
     CFS_ACCOUNT_DESCRIPTION = _get_config('CFS_ACCOUNT_DESCRIPTION', default='BCR')
     CFS_INVOICE_PREFIX = os.getenv('CFS_INVOICE_PREFIX', 'REG')
     CFS_RECEIPT_PREFIX = os.getenv('CFS_RECEIPT_PREFIX', 'RCPT')
+    CFS_PARTY_PREFIX = os.getenv('CFS_RECEIPT_PREFIX', '')
```
sumesh-aot, line 108: need fix on the env var name. Also better to default it to BCR? But check with John or Jyoti first.

pay-api/src/pay_api/services/cfs_service.py
sumesh-aot: It was needed when we were using entity names (first version of CC) as the party name. No longer needed

pay-api/tests/unit/services/test_pad_service.py
```diff
@@ -30,7 +30,7 @@ def test_create_account(session):
         'bankTransitNumber': 222,
         'bankAccountNumber': 33333333
     }
-    account = pad_service.create_account(name='Test Account', contact_info={}, payment_info=payment_info)
+    account = pad_service.create_account(identifier='Test Account', contact_info={}, payment_info=payment_info)
```
sumesh-aot, line 33: May be change the identifer to account id, to not confuse ?

sumesh-aot (approved):

PR #825, opened 2021-11-24
Cancel inv cfs

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/9316

*Description of changes:*

### DIFF
--- a/pay-api/src/pay_api/services/cfs_service.py
+++ b/pay-api/src/pay_api/services/cfs_service.py
@@ -434,6 +434,33 @@ class CFSService(OAuthService):
         current_app.logger.debug('>Created CFS Invoice NSF Adjustment')
         return adjustment_response.json()
 
+    @classmethod
+    def adjust_invoice(cls, cfs_account: CfsAccountModel, inv_number: str, amount: float):
+        """Add adjustment to the invoice."""
+        current_app.logger.debug('>Creating Adjustment for Invoice: %s', inv_number)
+        access_token: str = CFSService.get_token().json().get('access_token')
+        cfs_base: str = current_app.config.get('CFS_BASE_URL')
+        adjustment_url = f'{cfs_base}/cfs/parties/{cfs_account.cfs_party}/accs/{cfs_account.cfs_account}/sites/' \
+                         f'{cfs_account.cfs_site}/invs/{inv_number}/adjs/'
+        current_app.logger.debug('Adjustment URL %s', adjustment_url)
+
+        adjustment = dict(
+            comment='Invoice cancellation',
+            lines=[
+                {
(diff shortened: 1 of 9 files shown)


pay-api/src/pay_api/services/cfs_service.py
```diff
@@ -426,6 +434,33 @@ def add_nsf_adjustment(cls, cfs_account: CfsAccountModel, inv_number: str, amoun
...
+        adjustment_url = f'{cfs_base}/cfs/parties/{cfs_account.cfs_party}/accs/{cfs_account.cfs_account}/sites/' \
+                         f'{cfs_account.cfs_site}/invs/{inv_number}/adjs/'
+        current_app.logger.debug('Adjustment URL %s', adjustment_url)
+
+        adjustment = dict(
+            comment='Invoice cancellation',
+            lines=[
+                {
+                    'line_number': '1',
+                    'adjustment_amount': str(amount),
```
sumesh-aot, line 452: Adjustment amount has to be how much amount needs to be deducted from the invoice. So if invoice amount is 100, then adjustment_amount should be `-100`.

pay-api/src/pay_api/services/cfs_service.py
```diff
@@ -426,6 +434,33 @@ def add_nsf_adjustment(cls, cfs_account: CfsAccountModel, inv_number: str, amoun
...
+                    'activity_name': 'BC Registries - Invoice cancellation'
```
sumesh-aot, line 453: activity_name is CFS_ADJ_ACTIVITY_NAME from constants file

pay-api/src/pay_api/services/internal_pay_service.py
```diff
@@ -109,6 +109,11 @@ def process_cfs_refund(self, invoice: InvoiceModel):
         if routing_slip := RoutingSlipModel.find_by_number(routing_slip_number):
             routing_slip.remaining_amount += decimal.Decimal(invoice.total)
             routing_slip.flush()
+            invoice.invoice_status_code = InvoiceStatus.REFUND_REQUESTED.value
+        else:
```
sumesh-aot, line 113: If it's not new FAS routing slip, we don't support refund. So it should be raising error, correct ?

pay-api/src/pay_api/services/cfs_service.py
saravanpa-aot: Done

pay-api/src/pay_api/services/internal_pay_service.py
saravanpa-aot: Good catch.Done

sumesh-aot (approved):

PR #911, opened 2022-04-25
11211 - Fix remaining_amount for NSF. 

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/11211

*Description of changes:*
Fix remaining_amount for NSF.
Also prevent linking NSF to parent. 


pay-api/src/pay_api/services/fas/routing_slip.py
```diff
@@ -385,11 +385,9 @@ def update(cls, rs_number: str, action: str, request_json: Dict[str, any], **kwa
             RoutingSlipStatusTransitionService.validate_possible_transitions(routing_slip, status)
             status = RoutingSlipStatusTransitionService.get_actual_status(status)
 
+            # Our routing_slips job will create an invoice (under transactions in the UI).
             if status == RoutingSlipStatus.NSF.value:
-                total_paid_to_reverse: float = 0
-                for rs in (routing_slip, *RoutingSlipModel.find_children(routing_slip.number)):
-                    total_paid_to_reverse += rs.total
-                routing_slip.remaining_amount = -total_paid_to_reverse
+                routing_slip.remaining_amount += -routing_slip.total
```
shabeeb-aot, line 390: what about the child routing slip?

shabeeb-aot (approved):

pay-api/src/pay_api/services/fas/routing_slip.py
seeker25: When routing slips are linked from child to parent the balance is transferred to the parent.
I've also put in a block in there that doesn't allow linking NSF -> Active. Only allowed to link Active -> NSF etc.

seeker25: Scenario 1  - Two active linked tickets - Parent set to NSF - no problem the parent carries the balance
Scenario 2 - Two active linked tickets - Child set to NSF - **Scenario I think we'd have to look at**
Scenario 3 - NSF, then linked afterwards by Active - Balance transferred from Active -> NSF
Scenario 4 - NSF linked to Active - can't happen, will lose NSF status because it changes to LINK, recent changes blocks this

shabeeb-aot (approved):

PR #919, opened 2022-04-29
11899 - CSO total and service fee fix

ritvick:
*Issue #:*
https://github.com/bcgov/entity/issues/11899

*Description of changes:*
1. For CSO team, the service fee has to be multiplied with quantity
2. For all cases the total of line item was not correct, previously total was assigned total_excluding_service_fee, changed it to actual total.


pwei1018 (commented):
Test case?

pwei1018 (approved):

PR #963, opened 2022-07-14
11991 - UXA feedback on ticket 9014 - pay API - BC Online Error Modal

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/11991

*Description of changes:*
Small logging fixes. 
Add index for payment_account_id.
Dependency update. 
Fix unit test migrations (broke this whoops). 
Exclude versions, where necessary, this will allow the schemas to be built without querying the version tables. 


pay-api/src/pay_api/models/account_fee.py
```diff
@@ -70,6 +70,6 @@ class Meta:  # pylint: disable=too-few-public-methods
         """Returns all the fields from the SQLAlchemy class."""
 
         model = AccountFee
-        exclude = ['service_fee']
+        exclude = ['versions', 'service_fee']
```
seeker25, line 73: We included versions in auth, and it would query the versions table. I've opted to just put them in each schema instead of BaseSchema here because there are way less instances of them.

ritvick (approved):

PR #977, opened 2022-08-19
13309 - $0 Totals stuck in Approved for EJV/PAD

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/13309

*Description of changes:*
Move distribution code back to being set on total = 0, because it's used for getting service fees for EJV payment processing.

Removing this on the job side as well, we'll need to do a data fix to slot in all of the distribution codes  for totals with $0 (at least for EJV). 


seeker25:
This requires a data fix as well.

ritvick (approved):

PR #1011, opened 2022-11-15
Add dynamic exclusion of versions field, similar to SBC-AUTH.

seeker25:
This is to facilitate being able to use jupyter notebooks with SBC-AUTH and SBC-PAY at the same time.

### DIFF
--- a/pay-api/src/pay_api/models/base_schema.py
+++ b/pay-api/src/pay_api/models/base_schema.py
@@ -21,6 +21,12 @@ from .db import ma
 class BaseSchema(ma.ModelSchema):  # pylint: disable=too-many-ancestors
     """Base Schema."""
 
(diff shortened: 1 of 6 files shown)


pay-api/src/pay_api/models/base_schema.py
```diff
@@ -21,6 +21,12 @@
 class BaseSchema(ma.ModelSchema):  # pylint: disable=too-many-ancestors
     """Base Schema."""
 
+    def __init__(self, *args, **kwargs):
```
seeker25, line 24: https://github.com/bcgov/sbc-auth/blob/b6fe708e6d48648290bff62aa7c218b4b9dda5bc/auth-api/src/auth_api/schemas/base_schema.py#L26

lekshmimallika-aot (approved):

PR #1016, opened 2022-11-19
14316 - Override staff bcol code for CSO fee.

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/14316

*Description of changes:*
 Add in a role to override staff bcol code. Supports CSO's use case of partial refunds. 


seeker25:
I think this will take some more review.. looking at BCOL API there is a condition in there to check if staff or not..

If staff it calls a certain wsdl method.. non staff a different wsdl method.

Jxio (approved):

seeker25:
This needs to change a bit, 
image

PR #1036, opened 2022-12-07
11314 - Part 2 - Update to jobs for VOID routing slips.

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/11314

*Description of changes:*
Update to jobs for VOID routing slips.


seeker25:
Requirements.txt needs to be updated, I'll figure out an easy way today to grab the commit hash from the targeted branch - so it worked similar to before (while still being compatible with your Makefile changes).

kialj876 (approved):

kialj876:
> Requirements.txt needs to be updated, I'll figure out an easy way today to grab the commit hash from the targeted branch - so it works similar to before (while still being compatible with your Makefile changes).

It's already grabbing from the commit hash so I think you can just update that once your other PR is merged?

seeker25:
@kialj876 I changed the makefile, very slightly, should provide us the hashes now while pointing at a branch.

jobs/payment-jobs/requirements.txt
```diff
@@ -1,26 +1,46 @@
+-e git+https://github.com/bcgov/sbc-common-components.git@dd41940a5465750bfed5c5fe1b06d3375ca13dcd#egg=sbc_common_components&subdirectory=python
```
seeker25, line 1: I read the order doesn't matter in this file, it builds a dependency graph I believe

PR #1052, opened 2022-12-14
Release changes using launch darkly plus minor tweaks

seeker25:
We're wanting to have three changes under feature flags. Two for CSO directly. One for the new PAD invoice refund statuses.

Also changes to the reverse receipt process.


pay-api/flags.json
```diff
@@ -2,6 +2,9 @@
     "flagValues": {
         "string-flag": "a string value",
         "bool-flag": true,
-        "integer-flag": 10
+        "integer-flag": 10,
+        "NEW_REFUNDED_STATUSES": true,
```
seeker25, line 6: in PROD (will need to setup):
NEW_REFUNDED_STATUSES = false 
CSO_BCOL_FIX = false
BAD_CSO_SERVICE_FEE = true

pay-api/src/pay_api/__init__.py
```diff
@@ -41,6 +41,7 @@
 def create_app(run_mode=os.getenv('FLASK_ENV', 'production')):
     """Return a configured Flask App using the Factory method."""
     app = Flask(__name__)
+    app.env = run_mode
```
seeker25, line 44: Had to add this, otherwise in test mode it wouldn't read from the config file.

pay-api/src/pay_api/services/cfs_service.py
```diff
@@ -367,6 +368,16 @@ def create_account_invoice(cls, transaction_number: str, line_items: List[Paymen
                                            invoice_payload)
         return invoice_response.json()
 
+    @classmethod
+    def _build_reversal_comment(cls, operation: ReverseOperation):
+        """Build the comment for the reversal."""
+        return {
```
seeker25, line 374: Put this in to make it a bit more robust. Going to confirm with CAS. 
Top two exist already. 

Correction, the idea was to increment `cas_version_suffix`, and append it to a routing slip number if it's > 1

EX.

`234054347`, `cas_version_suffix` incremented becomes `234054347R2`... `234054347R3` etc..

pay-api/src/pay_api/services/base_payment_system.py
```diff
@@ -151,13 +151,14 @@ def _release_payment(invoice: Invoice):
 
     @staticmethod
     def _refund_and_create_credit_memo(invoice: InvoiceModel):
+        from pay_api.services import flags  # pylint: disable=import-outside-toplevel
```
seeker25, line 154: If I didn't put the import here, ran into circular dependency troubles or flags not having `is_on`

pay-api/src/pay_api/services/bcol_service.py
```diff
@@ -49,6 +49,7 @@ def get_payment_system_code(self):
     def create_invoice(self, payment_account: PaymentAccount,  # pylint: disable=too-many-locals
                        line_items: [PaymentLineItem], invoice: Invoice, **kwargs) -> InvoiceReference:
         """Create Invoice in PayBC."""
+        from pay_api.services import flags  # pylint: disable=import-outside-toplevel
```
seeker25, line 52: If I didn't put the import here, ran into circular dependency troubles or flags not having `is_on`

pay-api/src/pay_api/services/fee_schedule.py
```diff
@@ -284,6 +284,7 @@ def find_by_corp_type_and_filing_type(  # pylint: disable=too-many-arguments
             **kwargs
     ):
         """Calculate fees for the filing by using the arguments."""
+        from pay_api.services import flags  # pylint: disable=import-outside-toplevel
```
seeker25, line 287: If I didn't put the import here, ran into circular dependency troubles or flags not having `is_on`

pay-api/src/pay_api/__init__.py
kialj876: that's weird. Was it not getting any values or just not setting the test config version of them?

kialj876: Oh was it just not getting ff values?

seeker25: I'd run the unit tests in the debugger, would always be set to PRODUCTION strangely

kialj876 (approved):
Looks right to me

seeker25:
Deploying this to DEV, just to do a smoke test.

pay-api/src/pay_api/services/flags.py
```diff
@@ -59,12 +60,6 @@ def init_app(self, app):
             client = ldclient_get()
 
             app.extensions['featureflags'] = client
-            app.teardown_appcontext(self.teardown)
-
-    def teardown(self, exception):  # pylint: disable=unused-argument; flask method signature
```
seeker25, line 64: This teardown needed to be removed, otherwise flags get stuck being cached.  Everytime it evaluates it hits this warning:

image

PR #1115, opened 2023-02-16
15257 - Fix statements serialization issue

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/15257

*Description of changes:*
Use cattrs instead of pydantic. Minor fix in migration for upgrading sqlalchemy in the future.

### DIFF
--- a/pay-api/docker-entrypoint.sh
+++ b/pay-api/docker-entrypoint.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 echo 'starting application'
-gunicorn --bind 0.0.0.0:8080 --timeout 60 --workers 3  wsgi:application
+gunicorn --bind 0.0.0.0:8080 --timeout 600 --workers 3  wsgi:application
--- a/pay-api/docker-entrypoint.sh
+++ b/pay-api/docker-entrypoint.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 echo 'starting application'
-gunicorn --bind 0.0.0.0:8080 --timeout 360 --workers 3  wsgi:application
+gunicorn --bind 0.0.0.0:8080 --timeout 90 --workers 3  wsgi:application
--- a/pay-api/docker-entrypoint.sh
+++ b/pay-api/docker-entrypoint.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 echo 'starting application'
-gunicorn --bind 0.0.0.0:8080 --timeout 600 --workers 3  wsgi:application
+gunicorn --bind 0.0.0.0:8080 --timeout 360 --workers 3  wsgi:application
--- a/pay-api/docker-entrypoint.sh
+++ b/pay-api/docker-entrypoint.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 echo 'starting application'
-gunicorn --bind 0.0.0.0:8080 --timeout 100 --workers 3  wsgi:application
+gunicorn --bind 0.0.0.0:8080 --config /opt/app-root/gunicorn_config.py  wsgi:application
--- a/pay-api/docker-entrypoint.sh
+++ b/pay-api/docker-entrypoint.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 echo 'starting application'
-gunicorn --bind 0.0.0.0:8080 --timeout 90 --workers 3  wsgi:application
+gunicorn --bind 0.0.0.0:8080 --timeout 100 --workers 3  wsgi:application
--- a/pay-api/migrations/versions/b6e28faea978_rename_tables_to_plurals.py
+++ b/pay-api/migrations/versions/b6e28faea978_rename_tables_to_plurals.py
@@ -74,7 +74,8 @@ def upgrade():
     conn = op.get_bind()
     inspector = Inspector.from_engine(conn)
     tables = inspector.get_table_names()
-    metadata = MetaData(conn, reflect=True)
+    metadata = MetaData(conn)
+    metadata.reflect()
 
     table: str
     for table in tables:
@@ -89,7 +90,8 @@ def downgrade():
     conn = op.get_bind()
     inspector = Inspector.from_engine(conn)
     tables = inspector.get_table_names()
-    metadata = MetaData(conn, reflect=True)
+    metadata = MetaData(conn)
+    metadata.reflect()
     table_mapping_reversed = {y: x for x, y in table_mapping.items()}
     table: str
     for table in tables:
--- a/pay-api/src/pay_api/models/invoice.py
+++ b/pay-api/src/pay_api/models/invoice.py
@@ -160,6 +160,7 @@ class InvoiceSchema(AuditSchema, BaseSchema):  # pylint: disable=too-many-ancest
     @post_dump
     def _clean_up(self, data, many):  # pylint: disable=unused-argument
         """Clean up attributes."""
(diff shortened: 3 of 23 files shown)


pay-api/src/pay_api/models/invoice.py
```diff
@@ -173,3 +178,39 @@ def _clean_up(self, data, many):  # pylint: disable=unused-argument
...
 
         return data
+
+
+@define
+class InvoiceSearchModel:  # pylint: disable=too-few-public-methods, too-many-instance-attributes
+    """Main schema used to serialize invoice searches, plus csv / pdf export."""
+
+    id: int
+    corp_type_code: str  # TODO probably a good idea to move this.
```
seeker25, line 188: This will require some looking around to fix, as there are many different paths to this object.

pay-api/migrations/versions/b6e28faea978_rename_tables_to_plurals.py
```diff
@@ -74,7 +74,8 @@ def upgrade():
     conn = op.get_bind()
     inspector = Inspector.from_engine(conn)
     tables = inspector.get_table_names()
-    metadata = MetaData(conn, reflect=True)
+    metadata = MetaData(conn)
+    metadata.reflect()
```
seeker25, line 78: SqlAlchemy fix when we upgrade. I tried upgrading to 1.4, but fell back after I realized it wasn't necessary to upgrade.

pay-api/src/pay_api/models/invoice.py
```diff
@@ -66,6 +70,7 @@ class Invoice(Audit):  # pylint: disable=too-many-instance-attributes
     receipts = relationship('Receipt', lazy='joined')
     payment_account = relationship('PaymentAccount', lazy='joined')
     references = relationship('InvoiceReference', lazy='joined')
+    corp_type = relationship('CorpType', foreign_keys=[corp_type_code], lazy='select', innerjoin=True)
```
seeker25, line 73: Needed to add this relationship, because we were looking at invoice-> payment_line_item -> fee_schedule -> corp_type before?

pay-api/src/pay_api/models/invoice.py
```diff
@@ -131,7 +136,7 @@ class Meta(BaseSchema.Meta):  # pylint: disable=too-few-public-methods
         """Returns all the fields from the SQLAlchemy class."""
 
         model = Invoice
-        exclude = []
+        exclude = ['corp_type']
```
seeker25, line 139: exclude the relationship corp_type, we don't want it showing up in our regular invoice schema.

pay-api/src/pay_api/models/invoice.py
```diff
@@ -173,3 +178,43 @@ def _clean_up(self, data, many):  # pylint: disable=unused-argument
...
+    product: str
+    invoice_number: str
+
+    @classmethod
+    def from_row(cls, row):
+        """From row is used so we don't tightly couple to our database class.
+
+        https://www.attrs.org/en/stable/init.html
+        """
+        return cls(id=row.id, corp_type_code=row.corp_type.code, corp_type=CorpTypeSearchModel.from_row(row.corp_type),
```
seeker25, line 212: image

I think it would have been nice if this was automagic

pay-api/src/pay_api/services/payment.py
```diff
@@ -352,22 +353,8 @@ def create_payment_report_details(cls, purchases: Tuple, data: Dict):  # pylint:
...
-                # All we need for CSV and PDF and search.
-                line_item_schema = PaymentLineItemSchema(only=('gst', 'description', 'pst'))
-                line_item_dict = line_item_schema.dump(payment_line_item)
-                line_item_dict['filing_type_code'] = payment_line_item.fee_schedule.filing_type_code
-                invoice['line_items'].append(line_item_dict)
-                invoice['product'] = payment_line_item.fee_schedule.corp_type.product
-            if len(invoice_dao.references) > 0:
-                invoice['invoice_number'] = invoice_dao.references[0].invoice_number
-            data['items'].append(invoice)
+            invoice_dict = Converter().unstructure(InvoiceSearchModel.from_row(invoice_dao))
```
seeker25, line 356: Much easier

pay-api/src/pay_api/utils/converter.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Converter(cattrs.Converter):
+    """Addon to cattr converter."""
+
+    def __init__(self):
+        """Initialize function, add in extra unstructure hooks."""
+        super().__init__()
+        # More from cattrs-extras/blob/master/src/cattrs_extras/converter.py
+        self.register_unstructure_hook(Decimal, self._unstructure_decimal)
```
seeker25, line 14: May need to extend this, I'll have to check the output versus Marshmellow for some datasets.

pay-api/src/pay_api/models/invoice.py
```diff
@@ -173,3 +178,43 @@ def _clean_up(self, data, many):  # pylint: disable=unused-argument
...
+                   created_on=row.created_on, total=row.total, paid=row.paid,
+                   service_fees=row.service_fees, folio_number=row.folio_number, created_name=row.created_name,
+                   details=row.details, payment_account=PaymentAccountSearchModel.from_row(row.payment_account),
+                   business_identifier=row.business_identifier, created_by=row.created_by,
+                   status_code=row.invoice_status_code, payment_method=row.payment_method_code,
+                   line_items=[PaymentLineItemSearchModel.from_row(x) for x in row.payment_line_items],
```
seeker25, line 218: Not sure if it's possible to have validators for these? What happens if payment_line_items is None for example?

seeker25:
Still needs to be tested, but seems to pass all the unit tests.

pay-api/docker-entrypoint.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 echo 'starting application'
-gunicorn --bind 0.0.0.0:8080 --timeout 60 --workers 3  wsgi:application
+gunicorn --bind 0.0.0.0:8080 --timeout 600 --workers 3  wsgi:application
```
seeker25, line 4: I had to crank this up, depending on what your filtering on can take a while

seeker25:
image

kialj876 (approved):
Nice work 👍 

One optional comment. If you want we can make that change afterwards as its currently not using the config file anyways

pay-api/docker-entrypoint.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 echo 'starting application'
-gunicorn --bind 0.0.0.0:8080 --timeout 60 --workers 3  wsgi:application
+gunicorn --bind 0.0.0.0:8080 --timeout 100 --workers 3  wsgi:application
```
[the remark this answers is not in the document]
seeker25: I don't think the variables are being used? 

image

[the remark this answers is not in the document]
seeker25: I don't see any references to those variables -^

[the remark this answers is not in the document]
seeker25: Ahh I see 
`CMD ["gunicorn", "--bind", "0.0.0.0:8080", "--config", "/opt/app-root/gunicorn_config.py", "wsgi:application"]`

seeker25:
@kialj876 Not sure what you mean it's not using that config file anyways?

PR #1137, opened 2023-03-06
Use gevent instead of sync for gunicorn workers, this grants us the a…

seeker25:
…bility to use gevent.spawn in code.

Note: 
On localhost, when using `flask run` it doesn't behave the same as going through `gunicorn` when using `gevent.spawn`. 


seeker25:
This should make it easy to do something like this in code (async calls):

image

seeker25:
Not working with `asyncio` publishing messages for the queues. Looking for a work around, alternatively upgrade Flask and use `aiohttp`.

seeker25:
Easier to just use `asyncio` for parallel requests. Quart is built off of `asyncio` as well, and hopefully one day they'll merge with Flask.

PR #1155, opened 2023-03-28
15719 - Skip EJV Disbursement / Payment on holidays

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/15719


pay-api/src/pay_api/services/fee_schedule.py
```diff
@@ -312,11 +311,6 @@ def find_by_corp_type_and_filing_type(  # pylint: disable=too-many-arguments
         # Set transaction fees
         fee_schedule.service_fees = FeeSchedule.calculate_service_fees(fee_schedule_dao, account_fee)
 
-        # Special case for CSO partner type which is different from normal flow
```
seeker25, line 315: Clean up

jobs/payment-jobs/logging.conf
```diff
@@ -17,6 +17,11 @@ handlers=console
 qualname=api
 propagate=0
 
+[logger_invoke_jobs]
```
seeker25, line 20: Had to add these, pretty sure alembic wipes them out

jobs/payment-jobs/utils/logger.py
```diff
@@ -26,6 +26,6 @@ def setup_logging(conf):
 
     if conf and path.isfile(conf):
         logging.config.fileConfig(conf)
```
seeker25, line 28: This is called before the alembic migrations, I added in a line to call it again after the migrations. That way the loggers get restored. 

Normally payment jobs doesn't run migrations.

PR #1159, opened 2023-04-17
Get rid of HOLIDAYS_LIST. Otherwise we have to redefine it every year

seeker25:
… which is fairly annoying. If someone forgets it causes a hassle with having to resubmit EJV files.

Waiting for https://github.com/dr-prodigy/python-holidays v0.23 to be published here:

https://pypi.org/project/holidays/

It includes truth and reconciliation day

"Union recognizes additional stats though, like Easter Monday and boxing day."

Payment jobs wont pass until I update the build requirements to point at this merge.


Jxio (approved):

seeker25:
image
 
+ Truth and reconciliation day with holidays 0.23

PR #1234, opened 2023-09-05
15687 - Pay remove restx

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/15687

*Description of changes:*
- remove restx from pay-api


pay-api/src/pay_api/resources/ops.py
Jxio, line 1: can ops.py move to ../v1/ ?

ochiu: The ops endpoint is part of the /ops route and not under /api/v1, so I kept it out of the v1 folder as it is not specific to a version.

seeker25:
Please move `git mv` to preserve the history of the files.

seeker25 (changes requested):
Please use `git mv` to preserve the history of the files. This is more important in PAY than AUTH.

seeker25:
After reviewing with Ody, `git mv` wasn't the issue. It's done in the commits

seeker25:
Another thing to note we discovered: when moving these files - better not to Squash and merge as is

seeker25 (approved):

PR #1274, opened 2023-10-04
17827 - initial TDI17 file retrieval setup

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/17827

*Description of changes:*
- initial SFT poller set up for EFT File retrieval


seeker25 (approved):
Looks good, we should probably up the version of the service?

PR #1275, opened 2023-10-04
17830 - EFT TDI17 Parser

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/17830

*Description of changes:*
- EFT TDI17 Parser with test cases and utils for generating file payload


queue_services/payment-reconciliations/src/reconciliations/eft/eft_base.py
```diff
@@ -0,0 +1,104 @@
...
+            self.add_error(EFTParseError(EFTError.INVALID_RECORD_TYPE, self.index))
+
+    def parse_decimal(self, value: str, error: EFTError) -> decimal:
+        """Try to parse decimal value from a string, return None if it fails and add an error."""
+        try:
+            # ends with blank or minus sign, handle the minus sign situation
+            if value.endswith('-'):
+                value = '-' + value[:-1]
+
+            result = decimal.Decimal(value)
```
seeker25, line 56: should probably add str(value) just in case here

queue_services/payment-reconciliations/src/reconciliations/eft/eft_record.py
```diff
@@ -0,0 +1,91 @@
...
+        # Confirm line length is valid, skip if it is not, but add to error array
+        if not self.is_valid_length():
+            self.add_error(EFTParseError(EFTError.INVALID_LINE_LENGTH))
+            return
+
+        # Confirm record type is as expected
+        self.record_type = self.content[0:1]
+        self.validate_record_type(EFTConstants.TRANSACTION_RECORD_TYPE.value)
+
+        self.ministry_code = self.content[1:3].strip()
```
seeker25, line 67: Could we include a link to the spec? Or a comment listing the spec? 

That way the next person knows where were building this from?

queue_services/payment-reconciliations/src/reconciliations/eft/eft_record.py
```diff
@@ -0,0 +1,91 @@
...
+        self.program_code = self.content[3:7].strip()
```
seeker25, line 68: Would recommend writing a function to split the data and `.strip()`

queue_services/payment-reconciliations/src/reconciliations/eft/eft_record.py
```diff
@@ -0,0 +1,91 @@
...
+        self.transaction_sequence = self.content[24:27].strip()
+
+        # We are expecting a BCROS account number here, it is required
+        self.transaction_description = self.content[27:67].strip()
+        if len(self.transaction_description) == 0:
+            self.add_error(EFTParseError(EFTError.BCROS_ACCOUNT_NUMBER_REQUIRED))
+
+        self.deposit_amount = self.parse_decimal(self.content[67:80], EFTError.INVALID_DEPOSIT_AMOUNT)
+        self.currency = self.get_currency(self.content[80:82]).strip()
+        self.exchange_adj_amount = self.parse_decimal(self.content[82:95], EFTError.INVALID_EXCHANGE_ADJ_AMOUNT)
```
seeker25, line 81: Why are some stripped and others not? just curious

ochiu: any monetary value actually has a space at the end to indicate a positive value, and a minus sign if it is negative

seeker25:
We should probably put in a unit test where we use a real file and it parses the expected result

ochiu:
> We should probably put in a unit test where we use a real file (non programmatically) and it parses the expected result
> 
> You're using factories to build it which could be incorrect in the first place
> 
> EG. some of the EJV unit tests do that

Sounds good I will do this.

queue_services/payment-reconciliations/src/reconciliations/eft/eft_base.py
seeker25: EG.

Also Decimal(10.00) works differently than Decimal('10.00')

queue_services/payment-reconciliations/src/reconciliations/eft/eft_record.py
ochiu: Added a comment with the spec in eft_base

queue_services/payment-reconciliations/src/reconciliations/eft/eft_record.py
ochiu: Introduced extract_value that splits the data and performs strip.

queue_services/payment-reconciliations/src/reconciliations/eft/eft_base.py
ochiu: Thanks for the insight. Updated.

ochiu:
I bumped the versions, for pay-api as I added models in relation to EFT processing and for this component.

queue_services/payment-reconciliations/src/reconciliations/eft/eft_base.py
```diff
@@ -11,6 +11,50 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# TDI17 File Specifications
```
seeker25, line 15: Perfect this helps a ton

seeker25 (approved):
Looks great, thank you!

PR #1281, opened 2023-10-16
17382 - Update the Wording based on PAD Receipts

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/17382

*Description of changes:*


Jxio:
> bcgov/entity#17382

yes, but the payment method text is from api request payload body, I need to update other repo which send the request I think.

ochiu:
> > [bcgov/entity#17382](https://github.com/bcgov/entity/issues/17382)
> 
> yes, but the payment method text is from api request payload body, I need to update other repo which send the request I think.

Got it! thanks for explaining.

ochiu (approved):

PR #1307, opened 2023-10-25
18099 - PAY API - endpoint to add EFT Allowed flag to an account

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/18099

*Description of changes:*

### DIFF
--- a/pay-api/src/pay_api/services/payment_account.py
+++ b/pay-api/src/pay_api/services/payment_account.py
@@ -29,6 +29,7 @@ from pay_api.models import Invoice as InvoiceModel
 from pay_api.models import PaymentAccount as PaymentAccountModel
 from pay_api.models import PaymentAccountSchema
 from pay_api.models import StatementSettings as StatementSettingsModel
+from pay_api.models.invoice import Invoice
 from pay_api.services.cfs_service import CFSService
 from pay_api.services.distribution_code import DistributionCode
 from pay_api.services.queue_publisher import publish_response
@@ -589,6 +590,10 @@ class PaymentAccount():  # pylint: disable=too-many-instance-attributes, too-man
 
         if account_fees := AccountFeeModel.find_by_account_id(self.id):
             d['accountFees'] = AccountFeeSchema().dump(account_fees, many=True)
+
+        paid_by_eft = Invoice.find_eft_invoices(self.auth_account_id)
+        if len(paid_by_eft) > 0:
+            d['eft_eligible'] = True
 
         return d
 
--- a/pay-api/src/pay_api/models/invoice.py
+++ b/pay-api/src/pay_api/models/invoice.py
@@ -32,7 +32,7 @@ from .audit import Audit, AuditSchema
 from .base_schema import BaseSchema
 from .db import db, ma
 from .invoice_reference import InvoiceReferenceSchema
-from .payment_account import PaymentAccountSchema, PaymentAccountSearchModel
+from .payment_account import PaymentAccount, PaymentAccountSchema, PaymentAccountSearchModel
 from .payment_line_item import PaymentLineItem, PaymentLineItemSchema
 from .receipt import ReceiptSchema
 from ..utils.util import current_local_time
@@ -173,6 +173,15 @@ class Invoice(Audit):  # pylint: disable=too-many-instance-attributes
             )
         )
 
+        return query.all()
+
+    @classmethod
+    def find_eft_invoices(cls, pay_account_id: str):
+        """Return invoices paid with eft payment method."""
+        query = db.session.query(Invoice) \
+            .join(PaymentAccount, PaymentAccount.id == Invoice.payment_account_id) \
+            .filter(Invoice.payment_method_code == PaymentMethod.EFT.value) \
+            .filter(PaymentAccount.auth_account_id == pay_account_id)
         return query.all()
 
 


seeker25:
Shouldn't this be an endpoint you can use to enable EFT on a user's account? Aka they can't choose EFT until someone goes into the backend and flips the switch via notebook?

seeker25:
@ochiu 
```
FAILED tests/integration/test_worker_queue.py::test_update_credit_payment - sqlalchemy.exc.StatementError: (builtins.KeyError) 'LEGISLATIVE_TIMEZONE'

```

ochiu:
> @ochiu
> 
> ```
> FAILED tests/integration/test_worker_queue.py::test_update_credit_payment - sqlalchemy.exc.StatementError: (builtins.KeyError) 'LEGISLATIVE_TIMEZONE'
> ```

@Jxio @seeker25 caused by a recent change to the invoice model to have a default for the overdue_date and it uses the current_local_time() function requiring the LEGISLATIVE_TIMEZONE.

@Jxio would you be able to add the below to the config.py for the event_listener in this PR please? If preferred I can also submit another PR. Just let me know thanks!

`LEGISLATIVE_TIMEZONE = os.getenv('LEGISLATIVE_TIMEZONE', 'America/Vancouver')`

seeker25:
I figured as much

Jxio:
@seeker25 
image
I probably got misunderstanding on this ticket. Is it not a flag just indicate if eft allowed? -> if an account had EFT invoices historically, then it allows to have EFT payment option?

seeker25:
Yeah but we don't want to enable eft for every client only the 300 in the list mostly.

queue_services/events-listener/src/events_listener/config.py
```diff
@@ -24,6 +24,7 @@
 
 from dotenv import find_dotenv, load_dotenv
 
+LEGISLATIVE_TIMEZONE = os.getenv('LEGISLATIVE_TIMEZONE', 'America/Vancouver')
```
ochiu, line 27: Should move it into the config class

L56 `class _Config():`

seeker25:
image
1. Add enable_eft to datamodel for payment_accounts
2. This flag should control if the EFT payment method can be switched to in the payment method screen
3. Case by case basis, so a route needs to be built that a notebook can be ran to enable EFT to peoples accounts, payload org_id + enableEFT: (true or false)
4. When this flag is added.. it should trigger a welcome email.. so it needs to fire down a queue message and call the account mailer

pay-api/migrations/versions/2023_10_26_2ef58b39cafc_add_eft_enable_to_payment_accounts.py
```diff
@@ -0,0 +1,26 @@
...
+# revision identifiers, used by Alembic.
+revision = '2ef58b39cafc'
+down_revision = '194cdd7cf986'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("set statement_timeout=20000;")
+    op.add_column('payment_accounts', sa.Column('eft_enable', sa.Boolean, nullable=True))
```
ochiu, line 21: We can probably default the new flag to false and make it not nullable.

pay-api/migrations/versions/2023_10_26_7d5231f2b9ac_update_eft_enable_payment_accounts.py
```diff
@@ -0,0 +1,33 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    """Update eft_enable in payment_accounts table."""
+    # eft_enable set to true when it has invoices paid with EFT/DIRECT_PAY payment method historically
+    op.execute("set statement_timeout=20000;")
+    conn = op.get_bind()
+    conn.execute('UPDATE payment_accounts \
```
ochiu, line 23: I don't think this is necessary, we want to enable specific accounts via notebook. It will also give us flexibility to enable a subset of accounts if we want and we want to send a welcome email when we enable it for accounts.

A new route/endpoint is beneficial as we can secure it differently than the existing payment account endpoints and have separate logic like the welcome email when we enable this flag.

pay-api/migrations/versions/2023_10_26_2ef58b39cafc_add_eft_enable_to_payment_accounts.py
Jxio: ok sounds good

seeker25 (approved):

pay-api/migrations/versions/2023_10_26_7d5231f2b9ac_update_eft_enable_payment_accounts.py
Jxio: I removed this file, it gonna handle by ticket 18100

seeker25:
lint + unit tests :(

pay-api/src/pay_api/services/payment_account.py
```diff
@@ -564,6 +577,7 @@ def asdict(self, **kwargs):
         # to handle PAD 3 day period..UI needs bank details even if PAD is not activated
         is_future_pad = (self.payment_method == PaymentMethod.DRAWDOWN.value) and (self._is_pad_in_pending_activation())
         show_cfs_details = is_ob_or_pad or is_future_pad
+        d['eft_enable'] = self._dao.eft_enable
```
seeker25, line 580: I don't think you need this?

seeker25:
Something else I noticed, what about the route?

pay-api/src/pay_api/services/payment_account.py
Jxio: I tested with postman, when the eft_enable is false, it doesn't show up. I'm removing `eft_enable = fields.Boolean(...)`

Jxio:
> Something else I noticed, what about the route?

a separate route? I don't think we need it. it can be passed with payment-accounts route:
image

seeker25:
That's the GET, we still need a PUT or PATCH route so we can SET `ENABLE_EFT`

pay-api/src/pay_api/services/payment_account.py
seeker25: That's fine, it's boolean in your model isn't it? FE sets it to false if it's undefined?

seeker25: if it doesn't show up, can just assume it's false? There might be a helper function that removes false and Nones from the payload

seeker25: image

seeker25: `eft_enable = fields.Boolean(data_key='eft_enable', required=True)`

seeker25: image

seeker25: Might be able to mess with something in here -^

pay-api/src/pay_api/services/payment_account.py
```diff
@@ -699,3 +712,12 @@ def delete_account(cls, auth_account_id: str) -> PaymentAccount:
         if pay_account.statement_notification_enabled:
             pay_account.statement_notification_enabled = False
             pay_account.save()
+
+    @classmethod
+    def enable_eft(cls, auth_account_id: str) -> PaymentAccount:
+        """Enable EFT on the payment account."""
+        current_app.logger.debug('<patch_account')
```
seeker25, line 719: This isn't part of patch account, this is enable_eft

pay-api/src/pay_api/resources/v1/account.py
```diff
@@ -78,6 +78,29 @@ def get_account(account_number: str):
...
+    current_app.logger.info('<patch_account')
+
+    # Check if user is authorized to perform this action
+    check_auth(business_identifier=None, account_id=account_number, one_of_roles=[EDIT_ROLE])
+    try:
+        response = PaymentAccountService.enable_eft(account_number)
+    except ServiceUnavailableException as exception:
+        return exception.response()
+
+    status = HTTPStatus.ACCEPTED \
```
seeker25, line 96: Why would this be ACCEPTED or OK?

Shouldn't it throw a 400 if the org doesn't exist?

pay-api/src/pay_api/resources/v1/account.py
```diff
@@ -78,6 +78,29 @@ def get_account(account_number: str):
...
 
+@bp.route('/<string:account_number>/eft', methods=['PATCH'])
+@cross_origin(origins='*')
+@_tracing.trace()
+@_jwt.has_one_of_roles([Role.SYSTEM.value])
+def patch_account(account_number: str):
+    """Enable eft for an account."""
+    current_app.logger.info('<patch_account')
+
+    # Check if user is authorized to perform this action
```
seeker25, line 89: redundant comment

pay-api/tests/unit/services/test_payment_account.py
```diff
@@ -244,3 +244,13 @@ def test_delete_account_failures(session):
         PaymentAccountService.delete_account(payload.get('accountId'))
 
     assert excinfo.value.code == Error.TRANSACTIONS_IN_PROGRESS.code
+
+
+def test_patch_account(session, payload):
+    """Assert that patch payment account works."""
+    pay_account: PaymentAccountService = PaymentAccountService.create(payload)
```
seeker25, line 251: `payload = get_pad_account_payload()`

PR #1321, opened 2023-11-08
18510 - Add extra step to set transactions to REFUNDED if billable = 'f'

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/18510

*Description of changes:*
Should fix: https://github.com/bcgov/entity/issues/17772

### DIFF
--- a/pay-api/src/pay_api/services/ejv_pay_service.py
+++ b/pay-api/src/pay_api/services/ejv_pay_service.py
@@ -65,7 +65,8 @@ class EjvPayService(PaymentSystemService, OAuthService):
         # Publish message to the queue with payment token, so that they can release records on their side.
         self._release_payment(invoice=invoice)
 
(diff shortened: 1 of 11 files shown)


pay-api/src/pay_api/services/ejv_pay_service.py
```diff
@@ -76,6 +77,8 @@ def process_cfs_refund(self, invoice: InvoiceModel) -> str:  # pylint:disable=un
             2.1 Return REFUND_REQUESTED
         """
         current_app.logger.info(f'Received JV refund for invoice {invoice.id}, {invoice.invoice_status_code}')
+        if not payment_account.billable:
```
seeker25, line 80: The only non-billable account we have is for EJV.

ochiu (approved):

PR #1327, opened 2023-11-23
18730 - Add invoice id to description, because it doesn't show up in flow thr…

seeker25:
…ough.

https://github.com/bcgov/entity/issues/18730


ochiu (approved):

ochiu:
@seeker25 I noticed a few of the statement tests were erroring. I re-ran the check this morning and it passed.

I did some investigation it seemed like maybe a date time comparison issue. I will take a look at the statement jobs as I am currently working in there already.

PR #1409, opened 2024-02-10
19760 - Modify PAY-API to support partial refunds off of an endpoint

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/19760

*Description of changes:*

### DIFF
--- a/pay-api/src/pay_api/services/direct_pay_service.py
+++ b/pay-api/src/pay_api/services/direct_pay_service.py
@@ -14,6 +14,7 @@
 """Service to manage Direct Pay PAYBC Payments."""
 import base64
 from urllib.parse import unquote_plus, urlencode
+from typing import Dict, List
 
 from dateutil import parser
 from flask import current_app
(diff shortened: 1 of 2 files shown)


pay-api/src/pay_api/services/direct_pay_service.py
```diff
@@ -234,19 +236,28 @@ def _get_refund_token(cls):
...
         invoice_reference = InvoiceReferenceModel.find_by_invoice_id_and_status(
             invoice.id, InvoiceReferenceStatus.COMPLETED.value)
-        # Future: Partial refund support - This is backwards compatible
-        # refundRevenue: [{
-        #        'lineNumber': 1,
-        #        'refundAmount': 50.00,
-        #    }]
+
+        refund_details = []
+        for line in refund_revenue:
```
seeker25, line 246: Don't worry about this part, I'll handle it is part of the PAYBC ticket. 

I think you need to look at what happens before, invoice status, etc

seeker25: So we might want to store the payload or a representation of the partial refund request. Otherwise all we have to work with right now is the refund amount, but that doesn't really explain line by line where the refund was coming from.

seeker25: Payment line item table, doesn't have any field for `refund` and no concept of payment status

PR #1415, opened 2024-02-15
19760 - initial partial refunds model persistence

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/19760
*Description of changes:*
- update refund schema - add optional refundRevenue
- persist refund partial records
- extend converter to support camel to snake case conversion of properties
- extend converter to handle enum values vs enum objects

### DIFF
--- a/pay-api/src/pay_api/models/refunds_partial.py
+++ b/pay-api/src/pay_api/models/refunds_partial.py
@@ -11,12 +11,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Model to handle all operations related to Payment Line Item."""
+"""Model to handle all operations related to Payment Line Item partial refunds."""
 
+from decimal import Decimal
+from attrs import define
 from sqlalchemy import ForeignKey
+
+from .payment_line_item import PaymentLineItem
 from .audit import Audit
 from .base_model import VersionedModel
 from .db import db
+from ..utils.enums import RefundsPartialType
 
 
 class RefundsPartial(Audit, VersionedModel):  # pylint: disable=too-many-instance-attributes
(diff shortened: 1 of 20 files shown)


pay-api/src/pay_api/services/refund.py
```diff
@@ -268,12 +271,43 @@ def create_refund(cls, invoice_id: int, request: Dict[str, str], **kwargs) -> Di
...
+            db.session.add(refund_line)
+
+    @staticmethod
+    def _get_partial_refund_lines(refund_revenue: List[Dict]) -> List[RefundPartialLine]:
+        """Convert Refund revenue data to a list of Partial Refund lines."""
+        if not refund_revenue:
+            return []
+
+        refund_lines = []
+        for refund in refund_revenue:
```
ochiu, line 305: Investigating an issue when using the converter where the payload is camel case but the data class is snake case.

Trying to update the converter to allow for a parameter to do a camel to snake case conversion when it is instantiated.

pay-api/src/pay_api/utils/converter.py
```diff
@@ -1,25 +1,76 @@
...
 import cattrs
+from attrs import fields, has
+from cattrs.gen import make_dict_structure_fn, make_dict_unstructure_fn, override
 
 
 class Converter(cattrs.Converter):
     """Addon to cattr converter."""
 
-    def __init__(self):
+    def __init__(self, camel_to_snake_case: bool = False, enum_to_value: bool = False):
```
ochiu, line 15: We should probably convert this into a data class for ConverterOptions if we start adding more configurations in the future

pay-api/src/pay_api/services/refund.py
ochiu: Updated converter to handle camel to snake conversion

seeker25 (approved):

pay-api/src/pay_api/models/refunds_partial.py
```diff
@@ -50,6 +52,19 @@ class RefundsPartial(Audit, VersionedModel):  # pylint: disable=too-many-instanc
     refund_amount = db.Column(db.Numeric(19, 2), nullable=False)
     refund_type = db.Column(db.String(50), nullable=True)
 
+    @classmethod
+    def find_by_invoice_id(cls, invoice_id: int):
+        """Return refund partials by invoice id."""
+        return db.session.query(RefundsPartial)\
+            .join(PaymentLineItem, PaymentLineItem.id == RefundsPartial.payment_line_item_id)\
+            .filter(PaymentLineItem.invoice_id == invoice_id).all()
```
seeker25, line 60: Probably better to put this in the service and keep the models barebones?

You reference to payment_line_item in the model, which could cause some pain of circular dependencies in the future

seeker25: I wonder if there is a way to reference to PaymentLineItem, without having to include it at the top somehow

ochiu: agreed, good call

PR #1459, opened 2024-04-04
20607 - Add in TZ conversion

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/20607


Jxio (approved):

seeker25:
Leaving CSO alone, they haven't complained, plus have access to API endpoint to check and have their own running system.

PR #1462, opened 2024-04-05
20588 - Add verification logic in some .setters to avoid unnecessary db actions

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/20588

*Description of changes:*


seeker25:
Ahh I see you moved the IF up one line.. interesting. I think I tried this before and it caused issues, but I did it on all of the files.

seeker25 (approved):

PR #1490, opened 2024-04-19
20454 - Pay-queue JWT Verification

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/20454
*Description of changes:*

### DIFF
--- a/pay-api/tests/utilities/base_test.py
+++ b/pay-api/tests/utilities/base_test.py
@@ -379,7 +379,7 @@ def factory_payment_account(payment_system_code: str = 'PAYBC', payment_method_c
                             name=None,
                             branch_name=None):
     """Return Factory."""
-    # Create a payment account
+    # Create a payment  account
     account = PaymentAccount(
         auth_account_id=auth_account_id,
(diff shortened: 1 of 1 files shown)


pay-queue/src/pay_queue/external/gcp_auth.py
```diff
@@ -13,28 +13,31 @@
...
             Request(session=session),
-            audience=current_app.config.get('PAY_SUB_AUDIENCE')
+            audience=current_app.config.get('PAY_SUB_AUDIENCE') or 'pay-dev'
         )
-    except Exception as e:  # TODO fix
-        msg = f'Invalid token: {e}\n'
-    finally:
-        return msg
+        # Check if the email is verified and matches the configured email
+        required_email = current_app.config.get('VERIFY_PUBSUB_EMAIL') or 'sa-queue@gtksf3-dev.iam.gserviceaccount.com'
```
ochiu, line 25: Is this default okay? if this isn't configured in an upper environment it will default to this value that is used for dev?

PR #1536, opened 2024-05-21
Small tweak use corp type for NRO.

seeker25:


pay-api/src/pay_api/utils/util.py
```diff
@@ -259,9 +259,9 @@ def cents_to_decimal(amount: int):
 
 def get_topic_for_corp_type(corp_type: str):
     """Return a topic to direct the queue message to."""
-    product_code = CodeService.find_code_value_by_type_and_code(Code.CORP_TYPE.value, corp_type).get('product')
-    if product_code == Product.NRO.value:
+    if corp_type == CorpType.NRO.value:
```
seeker25, line 262: Product code of NRO is BUSINESS

PR #1545, opened 2024-05-29
20719 - Logic for interim statement when change payment away from EFT

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/20719

*Description of changes:*

### DIFF
--- a/pay-api/src/pay_api/services/payment_account.py
+++ b/pay-api/src/pay_api/services/payment_account.py
@@ -377,15 +377,13 @@ class PaymentAccount():  # pylint: disable=too-many-instance-attributes, too-man
                 PaymentMethod.EFT.value not in {account.payment_method, target_payment_method}:
             return
 
-        if (account.payment_method == PaymentMethod.EFT.value and
-                target_payment_method != PaymentMethod.EFT.value):
-            Statement.generate_interim_statement(account.auth_account_id,
-                                                 StatementFrequency.PAYMENT_METHOD_CHANGED.value,
-                                                 False)
-            return
-
         # Payment method has changed between EFT and other payment methods
-        Statement.generate_interim_statement(account.auth_account_id, StatementFrequency.MONTHLY.value, True)
+        statement_frequency = (
+            StatementFrequency.MONTHLY.value
+            if account.payment_method == PaymentMethod.EFT.value
+            else StatementFrequency.default_frequency().value
+        )
+        Statement.generate_interim_statement(account.auth_account_id, statement_frequency)
 
     @classmethod
     def _check_and_update_statement_settings(cls, payment_account: PaymentAccountModel):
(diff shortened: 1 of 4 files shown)


pay-api/src/pay_api/services/payment_account.py
```diff
@@ -373,11 +373,19 @@ def create(cls, account_request: Dict[str, Any] = None, is_sandbox: bool = False
...
+        if account.payment_method == target_payment_method or \
+                PaymentMethod.EFT.value not in {account.payment_method, target_payment_method}:
             return
 
-        # Payment method has changed and is EFT
-        Statement.generate_interim_statement(account.auth_account_id, StatementFrequency.MONTHLY.value)
+        if (account.payment_method == PaymentMethod.EFT.value and
+                target_payment_method != PaymentMethod.EFT.value):
+            Statement.generate_interim_statement(account.auth_account_id,
+                                                 StatementFrequency.PAYMENT_METHOD_CHANGED.value,
```
ochiu, line 383: This probably isn't necessary and the value for StatementFrequency.PAYMENT_METHOD_CHANGED isn't really a frequency.

We can probably use the default in enums.py when switching to a new payment method that isn't EFT since we know that is monthly.
`def default_frequency():`

ochiu: Probably can simplify the conditional if that is the case as well

```
statement_frequency = StatementFrequency.MONTHLY.value if account.payment_method == PaymentMethod.EFT.value else StatementFrequency.get_default_frequency

Statement.generate_interim_statement(account.auth_account_id, statement_frequency)

```

pay-api/src/pay_api/services/statement.py
```diff
@@ -225,7 +225,7 @@ def populate_overdue_from_invoices(statements: List[StatementModel]):
         return statements
 
     @staticmethod
-    def generate_interim_statement(auth_account_id: str, new_frequency: str):
+    def generate_interim_statement(auth_account_id: str, new_frequency: str, change_frequency: bool):
```
ochiu, line 228: Shouldn't need this flag, just set with whatever new_frequency is passed in

ochiu:
Just a bit of context as this ticket requires some coordination with #18619 where the implementation of the interim statement outputs will be done as we are trying to share the same template and underlying queries if possible.

This ticket will just implement the generation of the interim statement.

pay-api/src/pay_api/services/payment_account.py
Jxio: how about in the statement table? what frequency suppose to save there? it shows "Payment Method Changed" under frequency at UI
image

ochiu: Currently a statement record will return the payment methods for the duration of the statement using history records.

https://github.com/bcgov/sbc-pay/blob/main/pay-api/src/pay_api/models/statement.py#L73

and in auth web, if there are multiple payment methods we know it has changed for this statement

https://github.com/bcgov/sbc-auth/blob/main/auth-web/src/components/auth/account-settings/statement/Statements.vue#L319

ochiu (approved):
Looks good thanks!

seeker25 (approved):

seeker25:
I hate to say it, but might be easier to rebase this off of main.. I see alot of conflicts in the feature branch

seeker25:
Please rebase this off of `main`

Jxio:
please refer to this PR: https://github.com/bcgov/sbc-pay/pull/1558

PR #1567, opened 2024-06-10
Add in gunicorn_config also don't use werkzeug

seeker25:

### DIFF
--- a/pay-queue/Dockerfile
+++ b/pay-queue/Dockerfile
@@ -84,4 +84,4 @@ USER web
 FROM development_build AS production_build
 COPY --chown=web:web . /code
 
-CMD [ "python", "/code/app.py" ]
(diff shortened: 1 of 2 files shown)


pay-queue/Dockerfile
```diff
@@ -84,4 +84,4 @@ USER web
 FROM development_build AS production_build
 COPY --chown=web:web . /code
 
-CMD [ "python", "/code/app.py" ]
+CMD gunicorn --bind 0.0.0.0:8080 --config /code/gunicorn_config.py app:app
```
seeker25, line 87: This is fine in the other branch:

https://github.com/bcgov/sbc-pay/blob/c47847e5ac4373f9a0dea38e756374995cc9d51d/pay-queue/Dockerfile#L36

PR #1600, opened 2024-07-04
22206 - Refactor payment account service so it's more dynamic and less boiler…

seeker25:
… plate we can reuse this for other services that are complicated, also this should set us up to be able to have multiple CFS accounts active at once.

*Issue #:*
https://github.com/bcgov/entity/issues/22206

*Description of changes:*

### DIFF
--- a/pay-api/src/pay_api/services/payment_account.py
+++ b/pay-api/src/pay_api/services/payment_account.py
@@ -102,6 +102,7 @@ class PaymentAccount():  # pylint: disable=too-many-instance-attributes, too-man
         # Prevent recursion by checking if the attribute name starts with '__' (private attribute)
         if name == '_PaymentAccount__dao':
             super().__setattr__(name, value)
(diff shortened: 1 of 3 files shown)


pay-api/src/pay_api/services/payment_account.py
```diff
@@ -61,284 +61,51 @@ class PaymentAccount():  # pylint: disable=too-many-instance-attributes, too-man
...
-    def eft_enable(self, value: bool):
-        """Set the eft_enable."""
-        if self._eft_enable != value:
-            self._eft_enable = value
-            self._dao.eft_enable = value
+    def __getattr__(self, name):
+        """Dynamic way of getting the properties from the DAO, anything not in __init__."""
+        if hasattr(self._dao, name):
+            return getattr(self._dao, name)
+        return None
```
seeker25, line 97: I should probably put a warning here

Jxio (approved):

PR #1603, opened 2024-07-05
Disbursement refund date fix

seeker25:


Jxio (approved):

pay-api/src/pay_api/services/refund.py
```diff
@@ -287,6 +287,10 @@ def create_refund(cls, invoice_id: int, request: Dict[str, str], **kwargs) -> Di
         # set invoice status
         invoice.invoice_status_code = invoice_status or InvoiceStatus.REFUND_REQUESTED.value
         invoice.refund = invoice.total  # no partial refund
+        if invoice.invoice_status_code in (InvoiceStatus.REFUNDED.value,
+                                           InvoiceStatus.CANCELLED.value,
+                                           InvoiceStatus.CREDITED.value):
+            invoice.refund_date = datetime.now(tz=timezone.utc)
```
ochiu, line 293: If we are setting invoice.refund above, should we not always set the refund_date here?

Are there cases we set the invoice.refund and wouldn't have a refund_date based on status?

seeker25: as discussed teams

PR #1604, opened 2024-07-05
Warning fixes, update deps

seeker25:


pay-api/src/pay_api/services/payment_account.py
```diff
@@ -97,8 +97,7 @@ def __getattr__(self, name):
         """Dynamic way of getting the properties from the DAO, anything not in __init__."""
         if hasattr(self._dao, name):
             return getattr(self._dao, name)
-        current_app.logger.warning('Attribute %s not found.', name)
-        return None
+        raise AttributeError('Attribute %s not found.' % name)
```
seeker25, line 100: I know this is more dangerous, but then it ensures we aren't using undefined attributes

Jxio (approved):

PR #1609, opened 2024-07-09
PAY-ADMIN - Fix for issues with cookies, also fix OIDC so it uses JWKS etc.

seeker25:

### DIFF
--- a/pay-admin/admin/config.py
+++ b/pay-admin/admin/config.py
@@ -19,7 +19,6 @@ All modules and lookups get their configuration from the Flask config,
 rather than reading environment variables directly or by accessing this configuration directly.
 """
 
-import json
 import os
 import sys
 
--- a/pay-admin/admin/config.py
+++ b/pay-admin/admin/config.py
@@ -19,6 +19,7 @@ All modules and lookups get their configuration from the Flask config,
 rather than reading environment variables directly or by accessing this configuration directly.
 """
 
+import json
 import os
 import sys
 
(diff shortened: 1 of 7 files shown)


pay-admin/admin/config.py
```diff
@@ -86,12 +87,18 @@ class _Config():  # pylint: disable=too-few-public-methods
 
     # Normal Keycloak parameters.
     OIDC_CLIENT_SECRETS = os.getenv('PAY_OIDC_CLIENT_SECRETS', 'secrets/keycloak.json')
+    with open(OIDC_CLIENT_SECRETS, 'r', encoding='utf-8') as file:
+        secrets = json.load(file)
+    JWT_OIDC_JWKS_URI = secrets['web']['jwks_uri']
```
seeker25, line 92: I've added these strings to DEV/TEST/PROD

pay-admin/admin/config.py
```diff
@@ -92,6 +93,7 @@ class _Config():  # pylint: disable=too-few-public-methods
     OIDC_ID_TOKEN_COOKIE_SECURE = os.getenv('PAY_OIDC_ID_TOKEN_COOKIE_SECURE', 'True').lower() == 'true'
 
     PREFERRED_URL_SCHEME = 'https'
+    SESSION_TYPE = 'filesystem'
```
seeker25, line 96: Uses smaller cookie, saves 250-500 of these on the drive

pay-admin/poetry.lock
```diff
@@ -811,6 +811,30 @@ files = [
 [package.dependencies]
 Flask = "*"
 
+[[package]]
+name = "flask-session"
```
seeker25, line 815: This library makes the cookie very small, saves most of the session on the disk

ochiu (approved):
Looks good. Thanks for the comments explaining the changes.

PR #1636, opened 2024-07-22
21535 - Update url in eft refund email

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/21535

*Description of changes:*
1. Update url in eft refund email
2. Re-create table eft_refund_email_list with auto-increment id

### DIFF
--- a/pay-api/migrations/versions/2024_07_12_f921d5e32835_add_expense_authority_email.py
+++ b/pay-api/migrations/versions/2024_07_12_f921d5e32835_add_expense_authority_email.py
@@ -22,6 +22,7 @@ def upgrade():
                     sa.Column('first_name', sa.String(25), nullable=True),
                     sa.Column('last_name', sa.String(25), nullable=True),
                     sa.Column('email', sa.String(25), nullable=False),
+                    sa.PrimaryKeyConstraint('id')
                     )
 def downgrade():
     op.drop_table('eft_refund_email_list')
--- a/pay-api/migrations/versions/2024_07_12_f921d5e32835_add_expense_authority_email.py
+++ b/pay-api/migrations/versions/2024_07_12_f921d5e32835_add_expense_authority_email.py
@@ -21,8 +21,7 @@ def upgrade():
                     sa.Column('id', sa.Integer(), autoincrement=True, nullable=False),
                     sa.Column('first_name', sa.String(25), nullable=True),
                     sa.Column('last_name', sa.String(25), nullable=True),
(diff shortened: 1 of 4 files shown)


pay-api/migrations/versions/2024_07_12_f921d5e32835_add_expense_authority_email.py
```diff
@@ -22,6 +22,7 @@ def upgrade():
                     sa.Column('first_name', sa.String(25), nullable=True),
                     sa.Column('last_name', sa.String(25), nullable=True),
                     sa.Column('email', sa.String(25), nullable=False),
+                    sa.PrimaryKeyConstraint('id')
```
ochiu, line 25: This migration would have run already. I think you will need a new one to make changes to add a PK constraint to id.

Jxio: fixed

pay-api/migrations/versions/2024_07_22_f9c15c7f29f5_.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def upgrade():
+    op.drop_table('eft_refund_email_list')
+
+    op.execute(CreateSequence(Sequence('eft_refund_email_list_id_seq')))
+
+    op.create_table('eft_refund_email_list',
+        sa.Column('id', sa.Integer(), 
+                  sa.Sequence('eft_refund_email_list_id_seq'),
```
ochiu, line 27: If you are going to use a sequence, the model should probably refer to it as well.

ochiu (approved):

pay-api/migrations/versions/2024_07_22_f9c15c7f29f5_.py
```diff
@@ -0,0 +1,46 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'f9c15c7f29f5'
+down_revision = 'fb59bf68146d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.drop_table('eft_refund_email_list')
```
seeker25, line 21: probably not the best practice, if someone was using this table? but should be fine

seeker25 (approved):

pay-api/migrations/versions/2024_07_22_f9c15c7f29f5_.py
Jxio: this table is not deployed yet, so it's empty

PR #1648, opened 2024-07-26
21535 - Eft refund for eft_credits

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/21535

*Description of changes:*


seeker25 (approved):

pay-api/src/pay_api/services/eft_service.py
```diff
@@ -121,14 +123,40 @@ def create_shortname_refund(cls, request: Dict[str, str], **kwargs) -> Dict[str,
...
         html_body = cls._render_email_body(shortname, amount, comment, shortname_id)
 
         send_email(recipients, subject, html_body, **kwargs)
         refund.save()
 
+    @classmethod
+    def _refund_eft_credits(cls, shortname_id: str, amount: str):
+        """Refund the amount to eft_credits table based on short_name_id."""
+        refund_amount = Decimal(amount)
+        eft_credits = EFTCreditModel.find_by_short_name_id(int(shortname_id))
```
ochiu, line 139: EFTShortnames service has a couple of functions that might be useful
`def get_eft_credit_balance(short_name_id: int) -> Decimal:`
`def get_eft_credits(short_name_id: int) -> List[EFTCreditModel]:`

I would add validation to check if the refund amount is greater than the credit balance currently and throw business exception

pay-api/src/pay_api/services/eft_service.py
```diff
@@ -121,14 +123,40 @@ def create_shortname_refund(cls, request: Dict[str, str], **kwargs) -> Dict[str,
...
+                credit.remaining_amount -= refund_amount
+                refund_amount = Decimal(0)
+            else:
+                refund_amount -= credit_amount
+                credit.remaining_amount = Decimal(0)
+
+            credit.save()
+
+        if refund_amount > 0:
+            current_app.logger.error(f'Shortname {shortname_id} Refund amount exceed eft_credits remaining amount.')
```
ochiu, line 158: At this point you may have already done a save() even though there is not enough credits. We should probably roll back if conditions aren't met? My suggestion for the validation above should prevent this.

Should add a test for the business exception

pay-api/src/pay_api/services/eft_service.py
```diff
@@ -121,14 +126,39 @@ def create_shortname_refund(cls, request: Dict[str, str], **kwargs) -> Dict[str,
...
+    @classmethod
+    def _refund_eft_credits(cls, shortname_id: int, amount: str):
+        """Refund the amount to eft_credits table based on short_name_id."""
+        refund_amount = Decimal(amount)
+        eft_credits: List[EFTCreditModel] = EFTShortnames.get_eft_credits(shortname_id)
+        eft_credit_balance = EFTShortnames.get_eft_credit_balance(shortname_id)
+
+        if refund_amount > eft_credit_balance:
+            current_app.logger.error(f'Shortname {shortname_id} Refund amount exceed eft_credits remaining amount.')
+            raise BusinessException(Error.INVALID_TRANSACTION)
```
ochiu, line 147: I personally would add a new Error for this case, but I will leave it up to you.

Do we need the logging if we are already raising an exception?

ochiu (approved):
Looks good, thanks

pay-api/src/pay_api/services/eft_service.py
```diff
@@ -121,14 +126,38 @@ def create_shortname_refund(cls, request: Dict[str, str], **kwargs) -> Dict[str,
...
         html_body = cls._render_email_body(shortname, amount, comment, shortname_id)
 
         send_email(recipients, subject, html_body, **kwargs)
         refund.save()
 
+    @classmethod
+    def _refund_eft_credits(cls, shortname_id: int, amount: str):
+        """Refund the amount to eft_credits table based on short_name_id."""
+        refund_amount = Decimal(amount)
+        eft_credits: List[EFTCreditModel] = EFTShortnames.get_eft_credits(shortname_id)
```
seeker25, line 142: Low priority:

This typehint isn't needed, function already has the hint defined

PR #1666, opened 2024-08-06
Add in section for EFT CREATED -> to be returned as COMPLETED

seeker25:


ochiu (approved):

pay-api/src/pay_api/models/invoice.py
```diff
@@ -298,6 +303,7 @@ def from_row(cls, row):
         https://www.attrs.org/en/stable/init.html
         """
         # Similar to _clean_up in InvoiceSchema.
+        # In the future may need to add a mapping from EFT Status: CREATED -> COMPLETED
```
seeker25, line 306: Left this one out for now, CSO uses this route for reconciliation.. they might need to see if it's CREATED VS PAID

PR #1668, opened 2024-08-07
EFT - Create Invoice Clean up

ochiu:


jobs/payment-jobs/tasks/cfs_create_invoice_task.py
```diff
@@ -437,11 +437,6 @@ def _create_eft_invoices(cls):
                     current_app.logger.error(e)
                     continue
 
-            mailer.publish_mailer_events(QueueMessageTypes.EFT_INVOICE_CREATED.value, payment_account, {
```
ochiu, line 440: Looks like copied from PAD, we don't have this requirement for EFT and it isn't implemented on the account mailer. Removing we don't have unused queue messages published.

PR #1676, opened 2024-08-12
EFT Statement Report Fix

ochiu:
- remove unused input that is erroring


pay-api/src/pay_api/services/payment.py
```diff
@@ -467,7 +467,6 @@ def generate_payment_report(report_inputs: PaymentReportInput, **kwargs):  # pyl
 
             template_vars = {
                 'statementSummary': report_inputs.statement_summary,
-                'paymentTransactions': report_inputs.eft_transactions,
```
ochiu, line 470: Looked at the EFT report template in report-api. This currently does not seem to be used on the template. I have removed this input as it is causing errors.

seeker25 (approved):

PR #1682, opened 2024-08-14
22672 - Fix missing Approved invoices on get_invoices_owing_amount

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/22672

*Description of changes:*

### DIFF
--- a/pay-api/src/pay_api/services/statement.py
+++ b/pay-api/src/pay_api/services/statement.py
@@ -246,10 +246,9 @@ class Statement:  # pylint:disable=too-many-instance-attributes
                 .join(PaymentAccountModel, PaymentAccountModel.id == InvoiceModel.payment_account_id)
                 .filter(PaymentAccountModel.auth_account_id == auth_account_id)
                 .filter(InvoiceModel.invoice_status_code.in_((InvoiceStatus.SETTLEMENT_SCHEDULED.value,
-                                                              InvoiceStatus.PARTIAL.value,
-                                                              InvoiceStatus.CREATED.value,
                                                               InvoiceStatus.APPROVED.value,
                                                               InvoiceStatus.OVERDUE.value)))
+                .filter(InvoiceModel.payment_method_code == PaymentMethod.EFT.value)
                 .filter(~exists()
                         .where(StatementInvoicesModel.invoice_id == InvoiceModel.id))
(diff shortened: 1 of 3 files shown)


ochiu:
Thanks Jia, can we add / expand on an existing test to check on this invoices amount owing please?

pay-api/src/pay_api/services/statement.py
```diff
@@ -248,6 +248,7 @@ def get_invoices_owing_amount(auth_account_id: str):
                 .filter(InvoiceModel.invoice_status_code.in_((InvoiceStatus.SETTLEMENT_SCHEDULED.value,
                                                               InvoiceStatus.PARTIAL.value,
                                                               InvoiceStatus.CREATED.value,
```
seeker25, line 250: Take out CREATED, PARTIAL

Also shouldn't we filter for EFT here?

pay-api/src/pay_api/services/statement.py
```diff
@@ -248,6 +248,7 @@ def get_invoices_owing_amount(auth_account_id: str):
                 .filter(InvoiceModel.invoice_status_code.in_((InvoiceStatus.SETTLEMENT_SCHEDULED.value,
```
seeker25, line 248: SETTLEMENT_SCHED wont happen for EFT

seeker25, line 248: Do we need an EFT filter here?

pay-api/src/pay_api/services/statement.py
seeker25, line 250: I think we need to take out CREATED

seeker25: The parent function doesn't have CREATED:

image

pay-api/src/pay_api/services/statement.py
seeker25: I think we do

ochiu: yes, let's do that to make this more clean/safe. Only EFT uses the statement summary right now. We can consider expanding this by adding a payment method query parameter in the future if needed.

seeker25:
Associated this to the correct ticket 22672

seeker25 (approved):

PR #1686, opened 2024-08-15
22803 - Fix 65535+ query parameters error

seeker25:
https://github.com/bcgov/entity/issues/22803


jobs/payment-jobs/tests/jobs/test_generate_statements.py
```diff
@@ -274,6 +275,38 @@ def test_bcol_monthly_to_eft_statement(session):
...
+    invoice = factory_invoice(account)
+    statement_list = []
+    for _ in range(0, 70000):
+        statement_list.append({'created_on': '2024-01-01',
+                               'from_date': '2024-01-01 08:00:00',
+                               'to_date': '2024-01-04 08:00:00',
+                               'payment_account_id': f'{account.id}',
+                               'frequency': StatementFrequency.DAILY.value
+                               })
+    db.session.execute(insert(Statement), statement_list)
```
seeker25, line 296: faster suggested way than bulk insert according to docs

jobs/payment-jobs/tests/jobs/test_generate_statements.py
```diff
@@ -274,6 +275,38 @@ def test_bcol_monthly_to_eft_statement(session):
...
+    statement = db.session.query(Statement).first()
+    statement_invoices_list = []
+    for _ in range(0, 70000):
+        statement_invoices_list.append({
+            'statement_id': statement.id,
+            'invoice_id': invoice.id
+        })
+    db.session.execute(insert(StatementInvoices), statement_invoices_list)
+    StatementTask.generate_statements([datetime(2024, 1, 1, 8).strftime('%Y-%m-%d')])
+    assert True
```
seeker25, line 307: Usually throws exception before this

seeker25:
A better way of fixing this problem, I think would be rewriting it so we pass in the query object instead of a hard id list

ochiu (approved):
looks good to me

PR #1707, opened 2024-08-27
Clean out wire and unused code

seeker25:


pay-api/src/pay_api/services/cfs_service.py
```diff
@@ -492,7 +492,16 @@ def add_nsf_adjustment(cls, cfs_account: CfsAccountModel, inv_number: str, amoun
 
     @classmethod
     def adjust_invoice(cls, cfs_account: CfsAccountModel, inv_number: str, amount=0.0, adjustment_lines=None):
-        """Add adjustment to the invoice."""
+        """Add adjustment to the invoice.
```
ochiu, line 495: Thanks for this comment.

seeker25: We might need to use adjust invoice.. because we can adjust it for a reversal and adjust it back if we need to use the invoice again

ochiu (approved):
Nice clean up, thanks!

PR #1749, opened 2024-09-16
21570 - Email / CAS Supplier Number

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/21570

*Description of changes:*
- add short name columns for preferred email / cas supplier number
- add patch route to populate email / cas supplier number

### DIFF
--- a/pay-api/migrations/versions/2024_09_13_1fadb0e78d8c_21570_short_name_supplier_info.py
+++ b/pay-api/migrations/versions/2024_09_13_1fadb0e78d8c_21570_short_name_supplier_info.py
@@ -25,8 +25,16 @@ def upgrade():
         batch_op.add_column(sa.Column('cas_supplier_number', sa.String(25), nullable=True))
         batch_op.add_column(sa.Column('email', sa.String(100), nullable=True))
 
+    with op.batch_alter_table('eft_short_names_history', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('cas_supplier_number', sa.String(25), nullable=True))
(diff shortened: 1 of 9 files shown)


pay-api/migrations/versions/2024_09_13_1fadb0e78d8c_21570_short_name_supplier_info.py
```diff
@@ -0,0 +1,32 @@
...
+revision = '1fadb0e78d8c'
+down_revision = '2097573390f1'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table('eft_short_names', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('cas_supplier_number', sa.String(25), nullable=True))
+        batch_op.add_column(sa.Column('email', sa.String(100), nullable=True))
```
ochiu, line 26: Adding email and cas supplier number to short name.

As discussed with Ethan and @rodrigo-barraza these are essentially preferences to auto populate the EFT Refund form, but that form still will allow it to be edited so the fields will also remain on the eft_refunds table. This also allows for us to track what was 'APPROVED' at the time.

seeker25: what is the email used for?

ochiu: Short name contact email. Basically the main client email that manages the short name itself.

PR #1754, opened 2024-09-17
Clean up

seeker25:


pay-api/src/pay_api/resources/ops.py
```diff
@@ -38,5 +38,8 @@ def get_ops_healthz():
 @bp.route('readyz')
 def get_ops_readyz():
     """Return a JSON object that identifies if the service is setupAnd ready to work."""
-    # TODO: add a poll to the DB when called
+    try:
+        db.session.execute(SQL)
```
seeker25, line 42: Only impactful change

PR #1757, opened 2024-09-19
21539 - Short name refund history support

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/21539

*Description of changes:*
- Updates to support short name refund history

### DIFF
--- a/pay-api/src/pay_api/services/eft_service.py
+++ b/pay-api/src/pay_api/services/eft_service.py
@@ -162,15 +162,23 @@ class EftService(DepositService):
 
         current_app.logger.debug(f'Starting shortname refund : {shortname_id}')
 
-        refund = EftService._create_refund_model(request, shortname_id, amount, comment)
+        refund = EftService._create_refund_model(request, shortname_id, amount, comment).flush()
         EftService._refund_eft_credits(int(shortname_id), amount)
+
+        EFTHistoryService.create_shortname_refund(
+            EFTHistory(short_name_id=shortname_id,
+                       amount=amount,
+                       credit_balance=EFTCreditModel.get_eft_credit_balance(int(shortname_id)),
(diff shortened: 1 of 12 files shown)


pay-api/src/pay_api/services/eft_service.py
```diff
@@ -162,15 +162,23 @@ def create_shortname_refund(request: Dict[str, str], **kwargs) -> Dict[str, str]
 
         current_app.logger.debug(f'Starting shortname refund : {shortname_id}')
 
-        refund = EftService._create_refund_model(request, shortname_id, amount, comment)
+        refund = EftService._create_refund_model(request, shortname_id, amount, comment).flush()
         EftService._refund_eft_credits(int(shortname_id), amount)
 
+        EFTHistoryService.create_shortname_refund(
```
ochiu, line 168: When implementing approve/decline we need to find this history record by eft_refund_id and update the status.

For decline, we will need to return the refund amount back to the short name as well

pay-api/tests/unit/services/test_eft_short_name_historical.py
```diff
@@ -121,6 +126,76 @@ def test_create_statement_reverse(session, staff_user_mock):
...
         assert historical_record.related_group_link_id == 1
         assert historical_record.short_name_id == short_name.id
+        assert historical_record.eft_refund_id is None
+        assert historical_record.invoice_id is None
         assert historical_record.statement_number == 1234567
         assert historical_record.transaction_date == transaction_date
         assert historical_record.transaction_type == EFTHistoricalTypes.STATEMENT_REVERSE.value
+
+
+def test_create_invoice_refund(session, staff_user_mock):
```
ochiu, line 136: previous missed test, added this in along with short name refund

seeker25 (approved):

PR #1827, opened 2024-11-19
24432 - Bottom of EFT statement dynamic amount owing should be immutable

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/24432

*Description of changes:*


seeker25:
This approach wont work, for PAD - payment_date could be a couple of days ahead of when the invoice was created

PR #1829, opened 2024-11-20
24432 - Bottom of EFT statement dynamic amount owing should be immutable

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/24432


Jxio (approved):

ochiu (approved):
LGTM. Glad we were able to use dates to calculate on the fly and the backwards compatibility on non EFT statement calculations were kept intact.

PR #1846, opened 2024-12-03
Revert db.session.query

seeker25:


pay-api/src/pay_api/models/payment.py
```diff
@@ -319,9 +319,7 @@ def search_purchase_history(  # noqa:E501; pylint:disable=too-many-arguments, to
         if not return_all:
             count_future = executor.submit(cls.get_count, auth_account_id, search_filter)
             sub_query = cls.generate_subquery(auth_account_id, search_filter, limit, page)
-            # Need to do it this way otherwise doesn't know how to join for the subquery for some reason.
-            invoice_ids = {item for t in sub_query.all() for item in t}
```
seeker25, line 323: Doesn't work with > 65535 rows

PR #1862, opened 2024-12-13
Increase from 60k to 100k on CSV export

seeker25:


seeker25:
Need to do some quick testing for this

seeker25:
Highest counts are about 50k transactions in a month, 60k is close to the limit, 100k should be good

ochiu (approved):

PR #1873, opened 2025-01-10
24174- Handling statement template invoices with no payment date

rodrigo-barraza:
*Issue #:*
https://github.com/bcgov/entity/issues/24174

*Description of changes:*
Fixed an issue with the template not generating when there is no payment date on an invoice.

### DIFF
--- a/pay-api/src/pay_api/services/statement.py
+++ b/pay-api/src/pay_api/services/statement.py
@@ -358,9 +358,10 @@ class Statement:  # pylint:disable=too-many-public-methods
 
         latest_payment_date = None
         for invoice in statement_invoices:
-            if invoice.payment_date is not None:
-                if latest_payment_date is None or invoice.payment_date > latest_payment_date:
-                    latest_payment_date = invoice.payment_date
+            if invoice.payment_date is None:
+                continue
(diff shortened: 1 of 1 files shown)


pay-api/src/pay_api/services/statement.py
```diff
@@ -358,8 +358,9 @@ def _populate_statement_summary(cls, statement: StatementModel, statement_invoic
 
         latest_payment_date = None
         for invoice in statement_invoices:
-            if latest_payment_date is None or invoice.payment_date > latest_payment_date:
-                latest_payment_date = invoice.payment_date
+            if invoice.payment_date is not None:
```
ochiu, line 361: Less nesting if just continue if invoice.payment_date isn't there

if Invoice.payment_date is None:
   continue

ochiu (approved):

pay-api/src/pay_api/services/statement.py
seeker25:
```
latest_payment_date = max(
    (invoice.payment_date for invoice in statement_invoices if invoice.payment_date is not None),
    default=None
)
```

seeker25 (approved):

PR #1898, opened 2025-02-11
Set PRODUCT for BTR so queue messages may flow

seeker25:


kialj876 (approved):
Thanks!

PR #1941, opened 2025-04-09
27109 - Allow sandbox users to use account-ui to update their pad details

seeker25:
*Description of changes:*
Fix sandbox code so sandbox users can use account-ui to update their pad details


pay-api/src/pay_api/services/base_payment_system.py
```diff
@@ -355,9 +354,8 @@ def skip_invoice_for_sandbox(function):
     @functools.wraps(function)
     def wrapper(*func_args, **func_kwargs):
         """Complete any post invoice activities if needed."""
-        user: UserContext = func_kwargs["user"]
-        if user.is_sandbox():
```
seeker25, line 359: This isn't very useful when users use the GUI to create accounts or invoices etc, so we need to move it to an environment name check

pay-api/src/pay_api/resources/v1/bank_accounts.py
```diff
@@ -38,14 +38,28 @@ def post_bank_account_validate():
...
+                "bankNumber": request_json["bankNumber"],
+                "branchNumber": request_json["branchNumber"],
+                "isValid": True,
+                "message": ["VALID"],
+                "statusCode": 200,
+                "transitAddress": "",
+            }
+            status = HTTPStatus.OK
+        else:
+            response, status = CFSService.validate_bank_account(request_json), HTTPStatus.OK
```
seeker25, line 61: This wont work from GCP

PR #1946, opened 2025-04-23
25980 - Send feedback to client that credit is available on acct

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/25980
*Description of changes:*

### DIFF
--- a/pay-api/src/pay_api/templates/credit_add_notification.html
+++ b/pay-api/src/pay_api/templates/credit_add_notification.html
@@ -1,0 +1,15 @@
+# ${{ amount }} credit added to your account
+
+We've added ${{ amount }} credit to your account:
+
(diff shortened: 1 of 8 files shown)


pay-api/src/pay_api/templates/credit_add_notification.html
```diff
@@ -0,0 +1,15 @@
+# ${{ amount }} credit added to your account
+
+We've added ${{ amount }} credit to your account:
```
seeker25, line 3: isn't this the same thing repeated?

seeker25: @Jxio

seeker25: image

PR #1980, opened 2025-05-23
24607 - Add in cron for jobs as part of GCP migration

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/24607


Jxio (approved):

jobs/payment-jobs/schedules/run_create_cfs_invoices.test.cron
```diff
@@ -0,0 +1 @@
+*/30 0-4,15-23 * * *
```
ochiu, line 1: swapped with create account

PR #1983, opened 2025-05-26
28382 price list updates

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/28382

*Description of changes:*
- configurable fee schedule to show on price list
- filter updates
- add variable fee attribute to response
- test updates


ochiu:
> can you check the responses for `show_on_pricelist` specifically the CSO route one?
> 
> We'll have to let them know if we're changing the payload.. if not you can just filter them out

Will take a look, if it shows, I will filter it out

ochiu:
Confirmed show_on_pricelist does not appear, no action needed

PR #2013, opened 2025-06-17
28662 - Allow logging to work via secret

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/28662

*Description of changes:*
Switch logging to dict, that way we can easily put in as an environment variable


seeker25:
Hmm, should probably do this for pay-queue, pay-admin too

PR #2038, opened 2025-07-10
History table unit test checking

seeker25:
Small unit tests that check the integrity of migrations ensure that the history table lines up with the regular tables


seeker25:
Tested by removing columns under a migration, it would error

PR #2054, opened 2025-07-28
19544 - EFT Partial Refund / Refund UI support

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/19544
*Description of changes:*
- Move as much validation logic into refund service as possible to reduce duplication
- Clean up refundable states. Partials and Full refunds logic checks were conflicting, some tests were set up with incorrect invoice states
- Explicit full refundable states
- enable EFT Partial Refund flag
- Surface payment method codes to allow for checking partial_refund flag on refund UI
- rename payment method helper functions to be more short and concise
- New more precise Error messsages


pay-api/src/pay_api/services/base_payment_system.py
```diff
@@ -308,7 +312,8 @@ def _refund_and_create_credit_memo(invoice: InvoiceModel, refund_partial: List[R
         payment_account.flush()
 
         try:
-            PaymentSystemService._send_credit_notification(payment_account, refund_amount)
+            if send_credit_notification:
+                PaymentSystemService._send_credit_notification(payment_account, refund_amount)
```
ochiu, line 316: EFT doesn't send a credit notification currently. It goes to the EFT unsettled amount that clients cannot see. They would see a short name refund notification which is a separate flow.

pay-api/src/pay_api/services/code.py
```diff
@@ -64,6 +65,9 @@ def find_code_values_by_type(cls, code_type: str):
             elif code_type == CodeValue.ROUTING_SLIP_STATUS.value:
                 codes_models = RoutingSlipStatusCode.find_all()
                 schema = RoutingSlipStatusCodeSchema()
+            elif code_type == CodeValue.PAYMENT_METHODS.value:
+                codes_models = PaymentMethodModel.find_all()
+                schema = PaymentMethodSchema()
```
ochiu, line 70: To support Refund UI check to see if partial_refund  is set

ochiu:
Depends on pay-api backend changes for payment method codes

ochiu:
> Moved back to inprogress from review, probably need a resync from main
Thanks.

resynced, updated migration order along with some fixes and test updates

PR #2066, opened 2025-08-06
28594 - Pricelist GST

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/28594

*Description of changes:*
- update price list query to calculate GST


pay-api/tests/utilities/base_test.py
```diff
@@ -1090,7 +1090,8 @@ def factory_fee_schedule_model(
     service_fee: FeeCode = None,
     variable=False,
     show_on_pricelist=False,
-):
+    enable_gst=False,
```
ochiu, line 1093: @seeker25 I saw this column was renamed, will rename this along with the copied function based on your WIP PR

seeker25 (approved):

seeker25:
As discussed, move over some of the crossmodel queries from the model and into the service, should be good after that. We'll definitely look at removing the boilerplate setup that's going in the services probably expose the DAO directly but that can be for later

pay-api/tests/unit/models/test_fee_schedule.py
```diff
@@ -190,104 +190,3 @@ def get_or_create_corp_type(corp_type_code: str, corp_description: str, session)
     if existing:
         return existing
     return CorpType(code=corp_type_code, description=corp_description)
-
-
-def test_get_fee_details_all_products(session):
```
ochiu, line 195: These queries have been moved to the service, bye bye model tests.

Already tested as API and Service tests

PR #2078, opened 2025-08-14
30181 - Backwards compat changes for GST

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/30181 

*Description of changes:*

### DIFF
--- a/pay-api/src/pay_api/utils/converter.py
+++ b/pay-api/src/pay_api/utils/converter.py
@@ -89,6 +89,13 @@ class Converter(cattrs.Converter):
         return obj.isoformat() if obj else None
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


pay-api/src/pay_api/utils/converter.py
```diff
@@ -95,6 +95,8 @@ def remove_nones(data: Dict[Any, Any]) -> Dict[str, Any]:
         for key, val in data.items():
             if isinstance(val, dict):
                 new_data[key] = Converter.remove_nones(val)
+            elif isinstance(val, list):
+                new_data[key] = [Converter.remove_nones(item) if isinstance(item, dict) else item for item in val]
```
seeker25, line 99: this can impact performance and effect other fields I think we were just getting rid of None at the invoice base level, but PLI for example needed this extra bit

ochiu (approved):

PR #2089, opened 2025-08-20
Revert "Remove gst calculations from AR computation"

seeker25:
Reverts bcgov/sbc-pay#2085


pay-api/src/pay_api/services/cfs_service.py
```diff
@@ -624,6 +647,13 @@ def _process_line_item(
             existing_line["unit_price"] += amount_value
             existing_line["distribution"][0]["amount"] += amount_value
 
+            if (
+                line_data.has_gst
+                and existing_line.get("tax_classification") == TAX_CLASSIFICATION_GST
```
seeker25, line 652: Will remove `tax_classification`

PR #2097, opened 2025-08-22
27061 - Increase PAY test env memory

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/27061

*Description of changes:*

GCP is showing "memory limit of 1024 MiB exceeded with 1042 MiB used."
I tested locally with a larger statement and the memory usage was around 800 MB. It seems GCP handles memory control a bit differently, so I'm trying to set the test environment the same as prod to see if the error still occurs.

### DIFF
--- a/pay-api/devops/gcp/clouddeploy.yaml
+++ b/pay-api/devops/gcp/clouddeploy.yaml
@@ -45,6 +45,8 @@ serialPipeline:
       container-name: "pay-api-test"
       cloudsql-instances: "gtksf3-test:northamerica-northeast1:pay-db-test"
       service-account: "sa-api@gtksf3-test.iam.gserviceaccount.com"
+      resources-cpu: 4000m
+      resources-memory: 8Gi
  - targetId: gtksf3-sandbox
    profiles: [sandbox]
    strategy:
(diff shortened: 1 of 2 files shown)


pay-api/devops/gcp/clouddeploy.yaml
```diff
@@ -45,7 +45,8 @@ serialPipeline:
       container-name: "pay-api-test"
       cloudsql-instances: "gtksf3-test:northamerica-northeast1:pay-db-test"
       service-account: "sa-api@gtksf3-test.iam.gserviceaccount.com"
-      resources-memory: 1Gi
+      resources-cpu: 4000m
+      resources-memory: 8Gi
```
seeker25, line 49: @bolyachevets is this ok for test? Jia is doing some large statement testing

seeker25: She's going to try 2GI first

bolyachevets: sure. i set to prod values, it will get overwritten to the values in clouddeploy.yaml next time you redeploy

PR #2182, opened 2025-12-11
Fix statement

seeker25:


pay-api/src/pay_api/services/statement.py
```diff
@@ -457,9 +457,10 @@ def get_statement_report(statement_id: str, content_type: str, **kwargs):
             results=result_items,
         )
 
-        summary = Statement._build_statement_summary_for_methods(statement_dao, statement_purchases)
-        if summary:
-            report_inputs.statement_summary = summary
+        if extension == "pdf":
```
seeker25: already fixed in main:

image

PR #2203, opened 2025-12-30
History row conflict concurrency issue

seeker25:

### DIFF
--- a/pay-api/src/pay_api/services/pad_service.py
+++ b/pay-api/src/pay_api/services/pad_service.py
@@ -135,7 +135,7 @@ class PadService(PaymentSystemService, CFSService):
             payment_account = PaymentAccountModel.find_by_id_for_update(payment_account.id)
             pad_account_credit = payment_account.pad_credit or 0
             payment_account.pad_credit = 0 if pad_account_credit < invoice.total else pad_account_credit - invoice.total
-            payment_account.flush()
+            payment_account.save()
 
     @user_context
     @skip_complete_post_invoice_for_sandbox


Jxio (approved):

pay-api/src/pay_api/models/base_model.py
```diff
@@ -85,5 +85,7 @@ def find_by_id(cls, identifier: int) -> Self:
     def find_by_id_for_update(cls, identifier: int) -> Self:
         """Return model by id with lock to avoid race conditions."""
         if identifier:
-            return cls.query.filter_by(id=identifier).with_for_update().one_or_none()
+            query = cls.query.filter_by(id=identifier).with_for_update()
+            query = query.populate_existing()
```
seeker25: If you have an existing fetch of eg `payment_account.find_by_id`, this will refresh the row, if you don't do this you may have old data which causes a row conflict, we didn't see this in the versioned unit test because it only grabbed the data once

pay-api/src/pay_api/services/base_payment_system.py
```diff
@@ -306,7 +306,7 @@ def _refund_and_create_credit_memo(
             remaining_amount=refund_amount,
             account_id=invoice.payment_account_id,
             created_invoice_id=invoice.id,
-        ).flush()
+        ).save()
```
seeker25: This should be save, because the credit is already created in CMS

I was running into weird issues with `flush` to do with deadlocks  it would trigger a large query it seemed like, this should be save anyways

pay-api/src/pay_api/services/pad_service.py
```diff
@@ -135,7 +135,7 @@ def create_invoice(
             payment_account = PaymentAccountModel.find_by_id_for_update(payment_account.id)
             pad_account_credit = payment_account.pad_credit or 0
             payment_account.pad_credit = 0 if pad_account_credit < invoice.total else pad_account_credit - invoice.total
-            payment_account.flush()
+            payment_account.save()
```
seeker25: invoice creation commits after this anyways, should be ok

PR #2246, opened 2026-02-05
Add in is empty for statements

seeker25:

### DIFF
--- a/jobs/ftp-poller/poetry.lock
+++ b/jobs/ftp-poller/poetry.lock
@@ -2228,8 +2228,8 @@ werkzeug = "^3.1.4"
 [package.source]
 type = "git"
 url = "https://github.com/seeker25/sbc-pay.git"
-reference = "rename_direct_pay_direct_sale"
-resolved_reference = "e8e6973840de20de87d1f053ac3f801b2d0af47d"
+reference = "is_empty_statements"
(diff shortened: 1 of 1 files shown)


pay-api/migrations/versions/2026_02_04_is_empty_statements.py
```diff
@@ -0,0 +1,41 @@
...
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table("statements", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column(
+                "is_empty", sa.Boolean(), nullable=False, server_default="1"
+            )
+        )
```
seeker25: no index

pay-api/src/pay_api/models/statement.py
```diff
@@ -66,6 +67,7 @@ class Statement(BaseModel):
     payment_account_id = db.Column(db.Integer, ForeignKey("payment_accounts.id"), nullable=True, index=True)
     from_date = db.Column(db.Date, default=None, nullable=False)
     to_date = db.Column(db.Date, default=None, nullable=True)
+    is_empty = db.Column(db.Boolean(), nullable=False, default=True)
```
seeker25: no index let me know if you think this is a good idea or not @ochiu. I think this would be faster than joining to see statements_invoices is empty and could filter out alot of rows?

or is there a composite index we could use?

PR #2265, opened 2026-02-23
Remove unused gcp key

panish16:
*Issue #:*
https://github.com/bcgov/entity/issues/<Put the github issue number here>

*Description of changes:* Remove unused gcp key

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the sbc-pay license (Apache 2.0).

PR #2370, opened 2026-04-24
[DRAFT] fix: use SA impersonation for PubSub publishing in pay-api and payment-jobs

panish16:
*Issue #:*

*Description of changes:*

 ## Problem
PR #2265 removed `AUTHPAY_GCP_AUTH_KEY` from all service configs. When `payment-jobs` was deployed pointing to a revision that includes that removal, PubSub publishing started failing with:

403 PermissionDenied: pubsub.topics.publish on topic account-mailer-prod

The fallback is Application Default Credentials (`sa-job`), which is scoped only to Cloud Run execution and does not have `pubsub.publisher`.

  ## Solution
  Implement Service Account Impersonation so `sa-job` impersonates `sa-pubsub` (which holds `pubsub.publisher`) when publishing to topics.

  **Code changes:**
  - `pay-api/src/pay_api/services/gcp_queue/__init__.py` — subclass `GcpQueue`, override `init_app` to swap in a `PublisherClient` backed by `impersonated_credentials` targeting `AUTHPAY_SERVICE_ACCOUNT`
  - `pay-api/src/pay_api/config.py` + `jobs/payment-jobs/config.py` — expose `AUTHPAY_SERVICE_ACCOUNT` from env
  - `pay-api/devops/vaults.gcp.env` + `jobs/payment-jobs/devops/vaults.gcp.env` — add 1Password vault reference for `AUTHPAY_SERVICE_ACCOUNT` (matches existing `pay-queue` pattern)

  **Terraform prerequisite (bcregistry-sre):**
  `roles/iam.serviceAccountTokenCreator` must be granted on `sa-pubsub` to `sa-job@c4hnrd-prod` before this lands in prod, or publishing continues to fail. That binding is tracked in a separate bcregistry-sre PR.

  **Fallback safe:** when `AUTHPAY_SERVICE_ACCOUNT` is unset (local dev, tests), behaviour is identical to the base `GcpQueue` — no regression.


pay-api/src/pay_api/services/google_bucket_service.py
```diff
@@ -31,10 +31,12 @@ class GoogleBucketService:
...
         current_app.logger.info("Initializing Google Storage client.")
-        json_text = base64.b64decode(current_app.config.get("GOOGLE_STORAGE_SA")).decode("utf-8")
-        auth_json = json.loads(json_text, strict=False)
-        client = storage.Client.from_service_account_info(auth_json)
-        return client
+        google_storage_sa = current_app.config.get("GOOGLE_STORAGE_SA")
+        if google_storage_sa:
+            json_text = base64.b64decode(google_storage_sa).decode("utf-8")
+            auth_json = json.loads(json_text, strict=False)
+            return storage.Client.from_service_account_info(auth_json)
```
Jxio, line 38: should we allow it fallback to default when GOOGLE_STORAGE_SA is failed? like it existing but still failed.  what do you think @ochiu

pay-api/src/pay_api/services/gcp_queue/__init__.py
```diff
@@ -33,6 +33,28 @@
...
 """This module contains all the services used."""
 
+import google.auth
+from flask import Flask
 from gcp_queue import GcpQueue
+from google.auth import impersonated_credentials
+from google.cloud import pubsub_v1
 
-queue = GcpQueue()
+
```
bolyachevets, line 42: not sure we need the impersonation complexity. we can just grant sa-job relevant permissions

pay-api/src/pay_api/services/google_bucket_service.py
ochiu: I prefer it to just straight fail so it is obvious what is wrong instead of layers of things to figure out.

PR #2389, opened 2026-04-29
fix: read GOOGLE_STORAGE_SA from os.environ and use google.auth.default() fallback

panish16:
*Issue #:*

*Description of changes:* Travis removed GOOGLE_STORAGE_SA from config.py so current_app.config.get() always returns None. Read from os.environ first so the OCP secret value is picked up. Use google.auth.default() for the ADC/WIF fallback to supply both credentials and project, avoiding the storage client failing on OpenShift.

### DIFF
--- a/pay-api/src/pay_api/services/google_bucket_service.py
+++ b/pay-api/src/pay_api/services/google_bucket_service.py
@@ -18,6 +18,7 @@ import json
 import os
 import tempfile
 
+import google.auth
 from attrs import define
 from flask import current_app
 from google.cloud import storage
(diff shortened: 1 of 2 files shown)


pay-api/src/pay_api/services/google_bucket_service.py
```diff
@@ -31,12 +32,13 @@ class GoogleBucketService:
     def get_client() -> storage.Client:
         """Get a Google cloud storage client."""
         current_app.logger.info("Initializing Google Storage client.")
-        google_storage_sa = current_app.config.get("GOOGLE_STORAGE_SA")
+        google_storage_sa = os.environ.get("GOOGLE_STORAGE_SA") or current_app.config.get("GOOGLE_STORAGE_SA")
```
ochiu, line 35: Is os.environ.get() necessary? it was previously working with just current_app.config.get()?

ochiu:
@panish16 

From what I got from our recent conversations is that we want to remove GOOGLE_STORAGE_SA and it should use the default creds, which is why we made the current code changes to fall back on default.

Let me know if that isn't the case.

ochiu:
Seems it was a configuration issue and working in the feature branch. We can reopen if needed.

PR #2410, opened 2026-05-11
pay-admin : enable cloudsql connector

bolyachevets:
*Issue #:*

*Description of changes:*


pay-admin/admin/__init__.py
```diff
@@ -51,6 +51,25 @@ def create_app(run_mode=None):
     app.wsgi_app = ProxyFix(app.wsgi_app, x_proto=1)
     app.config.from_object(config.CONFIGURATION[run_mode])
 
+    # If Cloud SQL connection info is present, configure SQLAlchemy engine options
+    if app.config.get("CLOUDSQL_INSTANCE_CONNECTION_NAME"):
+        try:
+            from cloud_sql_connector import DBConfig
```
sumesh85: Same as the other PR, please move the import to module level if there is no specific reason to be inline.

pay-admin/admin/__init__.py
```diff
@@ -51,6 +51,25 @@ def create_app(run_mode=None):
...
+                database=app.config.get("DB_NAME", ""),
+                user=app.config.get("DB_USER", ""),
+                ip_type=app.config.get("DB_IP_TYPE"),
+                schema="public",
+                pool_timeout=30,
+                max_overflow=3,
+            )
+
+            app.config["SQLALCHEMY_ENGINE_OPTIONS"] = db_config.get_engine_options()
+        except Exception:
```
sumesh85: It's good practice to include the error details in log message,
Something like; 

```
except Exception as e:
   app.logger.exception("Failed to configure Cloud SQL DBConfig ", e)
```

sumesh85 (approved):
LGTM! Please check comments

pay-admin/admin/__init__.py
bolyachevets: @sumesh85  The reason to have inline import is mostly that not all code paths use it, but I am ok to move it to the top, as its not my codebase. Also, I will address the comments after the performance is verified, in case more changes need to be made.

pay-admin/admin/__init__.py
bolyachevets: @sumesh85 as above, I will address this after performance test. Might need to be smth like this to work in the logger: except Exception as e:
    app.logger.exception(f"Failed to configure Cloud SQL DBConfig: {e}")
    raise

PR #2421, opened 2026-05-25
Handle empty org contacts in get_account_info_with_contact

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/33381
*Description of changes:*

### DIFF
--- a/pay-api/src/pay_api/services/auth.py
+++ b/pay-api/src/pay_api/services/auth.py
@@ -213,5 +213,8 @@ def get_account_info_with_contact(**kwargs) -> dict:
 
         account_info = kwargs.get("auth").get("account")
         contacts = contact.get("contacts") or []
+        if not contacts:
+            current_app.logger.warning("No contacts found for account id %s", account_id)
+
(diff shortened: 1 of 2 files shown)


pay-api/src/pay_api/services/auth.py
```diff
@@ -212,5 +212,6 @@ def get_account_info_with_contact(**kwargs) -> dict:
         ).json()
 
         account_info = kwargs.get("auth").get("account")
-        account_info["contact"] = contact["contacts"][0]
+        contacts = contact.get("contacts") or []
+        account_info["contact"] = contacts[0] if contacts else {}
```
sumesh85, line 219: Can we add logger warning here if no contacts exist for the account; I guess contact is mandatory for accounts, the logger with account id would help us investigate the issue.

Jxio: added

sumesh85 (approved):
LGTM

ochiu (approved):

PR #2435, opened 2026-06-03
33384 - fix transaction 500 for NRO OB without invoice reference

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/33384

*Description of changes:*
invoice reference not existing because the flow of pay with credit card for NRO with online banking account. It created invoice reference in a patch request after user click "Pay", but it need login.


ochiu:
As discussed, this touches a few different flows. Update the ticket so regression testing can be done.

ochiu (approved):

PR #2439, opened 2026-06-04
33382 - fix(pay-api) - reject PAD bank updates while CFS account pending

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/33382

*Description of changes:*

### DIFF
--- a/pay-api/src/pay_api/services/pad_service.py
+++ b/pay-api/src/pay_api/services/pad_service.py
@@ -17,6 +17,7 @@ from typing import Any
 
 from flask import current_app
 
+from pay_api.exceptions import BusinessException
 from pay_api.models import CfsAccount as CfsAccountModel
 from pay_api.models import Invoice as InvoiceModel
 from pay_api.models import PaymentAccount as PaymentAccountModel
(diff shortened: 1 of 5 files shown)


pay-api/src/pay_api/services/pad_service.py
```diff
@@ -68,11 +70,17 @@ def create_account(
...
             str(payment_info.get("bankInstitutionNumber")) != cfs_account.bank_number
             or str(payment_info.get("bankTransitNumber")) != cfs_account.bank_branch_number
             or str(payment_info.get("bankAccountNumber")) != cfs_account.bank_account_number
-        ):
+        )
+        # CFS IDs are populated by the CREATE_CFS_ACCOUNTS job; reject bank updates until setup is complete
+        # to avoid calling CFS with None party/account/site values.
+        cfs_setup_incomplete = not (cfs_account.cfs_party and cfs_account.cfs_account and cfs_account.cfs_site)
+        if bank_info_changed and (cfs_setup_incomplete or cfs_account.status == CfsAccountStatus.PENDING.value):
+            raise BusinessException(Error.CFS_ACCOUNT_SETUP_IN_PROGRESS)
```
ochiu, line 82: Can we allow them to update it instead? I suspect this scenario is likely if a user made a mistake and wanted to update the information right away.

Jxio: I wasn’t sure we should allow bank updates while CFS setup is pending..

Jxio: I'm a bit worried that if a user updates bank details while the job is running on the same pending record(e.g. the job reads old bank info, or a save overwrites the user's latest values, make cfs out of sync).

ochiu: You are probably right, CFS can be slow sometimes and we don't want the update to happen during the create. Just confirm this error surfaces in the UI appropriately so the user knows to try again in a few minutes.

ochiu (approved):

PR #2489, opened 2026-06-09
33547 - transactions partial refund search fix

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/33547

*Description of changes:*
- Fix issue where partial invoice statuses were not properly filtering out declined refunds

### DIFF
--- a/pay-api/src/pay_api/services/invoice_search.py
+++ b/pay-api/src/pay_api/services/invoice_search.py
@@ -219,19 +219,34 @@ class InvoiceSearch:
         query = InvoiceService.filter_date(query, search_filter)
         return query
 
+    @staticmethod
+    def _partial_refund_with_approved_status_exists(is_credit: bool):
+        """Return an EXISTS clause for partial refunds with an approved refund status."""
+        return exists().where(
+            and_(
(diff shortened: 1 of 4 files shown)


pay-api/src/pay_api/services/invoice_search.py
```diff
@@ -219,19 +219,34 @@ def filter(cls, query, auth_account_id: str, search_filter: dict, include_joins=
...
         return query
 
+    @staticmethod
+    def _partial_refund_with_approved_status_exists(is_credit: bool):
+        """Return an EXISTS clause for partial refunds with an approved refund status."""
+        return exists().where(
+            and_(
+                RefundsPartial.invoice_id == Invoice.id,
+                RefundsPartial.is_credit.is_(is_credit),
+                exists().where(
```
ochiu, line 229: This was implemented before the refund flow was fully in place. All partial refunds will have a refund_id and we need to filter on the appropriate approval statuses since this can be declined and should not be returned in that case.

Jxio (approved):

sumesh85 (approved):

PR #2507, opened 2026-06-11
Tracing module update

sumesh85:
*Issue #:*

*Description of changes:*


ochiu (approved):

pay-api/poetry.lock
Jxio, line 1: we might need to get our poetry version fixed across apps, using different Poetry versions across the team might generate unnecessary diff noise in the lock file and may introduce inconsistencies in dependency resolution.

sumesh85: Thanks for raising this again. Do we have a standard version to use ? I can update mine to align.

Jxio: no we don't, I just noticed our app using different versions

PR #2508, opened 2026-06-11
33586 : Generalize non gov disbursement

sumesh85:
*Issue #:*
https://github.com/bcgov/entity/issues/33586

*Description of changes:*
- Generalize non-government partner AP disbursement from BCA-hardcoded to config-table-driven, supporting any number of non-gov partners                         
- Add non_gov_disbursement_config table (corp_type_code, cas_supplier_number, cas_supplier_site); migration gets default value from env vars                                                                                              
- Fixed disbursement amount calculation - excludes service fees
- Remove BCA_SUPPLIER_NUMBER/LOCATION config keys; supplier details now come from the DB config                      
                                                  


pay-api/migrations/versions/2026_06_09_add_non_gov_disbursement_config.py
```diff
@@ -0,0 +1,52 @@
...
+
+    non_gov_disbursement_config = table(
+        "non_gov_disbursement_config",
+        column("corp_type_code", sa.String),
+        column("cas_supplier_number", sa.String),
+        column("cas_supplier_site", sa.String),
+    )
+
+    bca_supplier_number = os.getenv("BCA_SUPPLIER_NUMBER", "")
+    bca_supplier_location = os.getenv("BCA_SUPPLIER_LOCATION", "")
```
Jxio: BCA_SUPPLIER_NUMBER & BCA_SUPPLIER_LOCATION not existing on pay-api I think

sumesh85: It exists for the job; however you have a good point that it is available for the job; not with the migration job. I will change this to hard coded values.

Jxio (approved):
lgtm, just before merge, have you had a chance to run the AP job locally against prod-copy data, compared with the current uploaded file from prod?

sumesh85:
> lgtm, just before merge, have you had a chance to run the AP job locally against prod-copy data, compared with the current uploaded file from prod?

I didn't do it; reason being the code change doesn't touch anything on the existing AP jobs; and it changes only how we were doing the BCA AP job. So wasn't seeing a value with that effort.

pay-api/src/pay_api/models/non_gov_disbursement_config.py
```diff
@@ -0,0 +1,37 @@
...
+
+    __mapper_args__ = {
+        "include_properties": [
+            "corp_type_code",
+            "cas_supplier_number",
+            "cas_supplier_site",
+        ]
+    }
+
+    corp_type_code = db.Column(db.String(10), ForeignKey("corp_types.code"), primary_key=True)
```
ochiu, line 42: Do we want at least a created_on? So we know when we put it in an environment?

ochiu: What about an enabled flag as well? Incase one of them has something wrong and we want to disable it without deleting the record

sumesh85: Yes, good idea, will add Audit and the flag.

pay-api/scripts/data/non_gov_disbursement_config_seed.sql
```diff
@@ -0,0 +1,7 @@
+-- Seed non_gov_disbursement_config for non-government partners paid via AP/EFT through CGI.
+-- Run once per environment after migration d4e5f6a7b8c9.
+-- Replace placeholders with the correct CAS supplier credentials for each environment.
+
+INSERT INTO non_gov_disbursement_config (corp_type_code, cas_supplier_number, cas_supplier_site, created_by)
```
ochiu, line 5: Thanks!

ochiu (approved):

PR #2556, opened 2026-06-17
33375 - cfs token refresh / caching

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/33375
https://github.com/bcgov/entity/issues/33380

*Description of changes:*
- add cfs token caching by key
- oauth service error logging, output request url

### DIFF
--- a/pay-api/src/pay_api/config.py
+++ b/pay-api/src/pay_api/config.py
@@ -137,6 +137,8 @@ class _Config:  # pylint: disable=too-few-public-methods
     CFS_INVOICE_PREFIX = os.getenv("CFS_INVOICE_PREFIX", "REG")
     CFS_RECEIPT_PREFIX = os.getenv("CFS_RECEIPT_PREFIX", "RCPT")
     CFS_PARTY_PREFIX = os.getenv("CFS_PARTY_PREFIX", "BCR-")
+    # Intentionally set lower to facilitate faster recovery from CFS outages / issues
+    CFS_TOKEN_CACHE_TIMEOUT = int(os.getenv("CFS_TOKEN_CACHE_TIMEOUT", "300"))
     PAY_CONNECTOR_AUTH = os.getenv("PAY_CONNECTOR_AUTH", "")
 
     # EFT Config
--- a/pay-api/src/pay_api/services/cfs_service.py
+++ b/pay-api/src/pay_api/services/cfs_service.py
@@ -15,13 +15,14 @@
 
 import base64
 import re
+import time
 from collections import defaultdict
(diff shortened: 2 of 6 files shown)


pay-api/src/pay_api/services/cfs_service.py
```diff
@@ -72,6 +73,9 @@ class ProcessingContext:
     negate: bool
 
 
+_token_cache: dict[str, tuple[Response, float]] = {}
```
ochiu, line 76: Opted for a custom dictionary so we can store it by payment system key, since we use two different CFS accounts. Originally I was going to use memoize similar to another service call but it limited the ability for us to define a different token lifespan from different services.

pay-api/src/pay_api/services/oauth_service.py
```diff
@@ -101,12 +101,12 @@ def post(  # pylint: disable=too-many-arguments
                 return response.iter_content()
 
         except (ReqConnectionError, ConnectTimeout) as exc:
-            current_app.logger.error("---Error on POST---")
+            current_app.logger.error(f"---Error on POST--- {endpoint}")
```
ochiu, line 104: Additional logging info for errors so we can see if there are higher occurences of issues on certain service calls

pay-api/src/pay_api/config.py
```diff
@@ -137,6 +137,8 @@ class _Config:  # pylint: disable=too-few-public-methods
     CFS_INVOICE_PREFIX = os.getenv("CFS_INVOICE_PREFIX", "REG")
     CFS_RECEIPT_PREFIX = os.getenv("CFS_RECEIPT_PREFIX", "RCPT")
     CFS_PARTY_PREFIX = os.getenv("CFS_PARTY_PREFIX", "BCR-")
+    # Intentionally set lower to facilitate faster recovery from CFS outages / issues
+    CFS_TOKEN_CACHE_TIMEOUT = int(os.getenv("CFS_TOKEN_CACHE_TIMEOUT", "300"))
```
ochiu, line 141: The default will be lower on pay-api so it can recover faster if there are CFS issues that are token related.
This will be set higher for jobs and potentially pay queue

sumesh85: Did you check the token expiry from CFS token now in prod ?

sumesh85 (approved):
LGTM! Just see my comment

pay-api/src/pay_api/config.py
ochiu: Yes currently it is max an hour and I confirmed it with a real token

PR #2558, opened 2026-06-17
poetry update, sync dependencies with main and define cfs token timeout

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/#33375

*Description of changes:*
- sync dependencies to main
- cfs token cache
- poetry update


Jxio (approved):

pay-api/src/pay_api/services/cfs_service.py
```diff
@@ -500,7 +500,7 @@ def get_token(payment_system=PaymentSystem.PAYBC):
             additional_headers={"Pay-Connector": current_app.config.get("PAY_CONNECTOR_AUTH")},
         )
         if token_response.json().get("access_token"):
-            timeout = current_app.config.get("CFS_TOKEN_CACHE_TIMEOUT")
+            timeout = current_app.config.get("CFS_TOKEN_CACHE_TIMEOUT", "300")
```
sumesh85, line 503: Do we need defaut here if the config is already defaulted ?

sumesh85 (approved):
LGTM! Please check a minor comment

pay-api/src/pay_api/services/cfs_service.py
ochiu: Originally, I didn't have it, but if anything else uses this and is missing the config it can just default to the 5 minutes.

PR #2583, opened 2026-07-16
34018: Add refund allowed flag for products

sumesh85:
*Issue #:*
https://github.com/bcgov/entity/issues/34018

*Description of changes:*


ochiu:
We will need to update invoice_search service query that surfaces refundable flags to the UI. It should allow the UI to automatically hide/disable refund buttons

https://github.com/bcgov/sbc-pay/blob/main/pay-api/src/pay_api/services/invoice_search.py#L126

pay-api/src/pay_api/models/search/invoice_composite_model.py
```diff
@@ -50,34 +50,46 @@ def get_latest_refund_status_expr():
     return select(latest_refund_subq.c.refund_status).scalar_subquery()
 
 
+def get_corp_type_refund_allowed_expr():
+    """Return an EXISTS clause asserting the invoice's corp type permits refunds."""
+    return (
+        select(CorpType.code)
```
ochiu, line 56: looks good, we can probably add a test for it.

API test - test_refund_approval should have some examples testing the refundable flags search and gets on the single invoice composite record

ochiu (approved):

pay-api/src/pay_api/models/search/invoice_composite_model.py
sumesh85: Added new test case

PR #2611, opened 2026-08-11
34304: Adding changes for PAD withhold

sumesh85:
*Issue #:*
https://github.com/bcgov/entity/issues/34304

*Description of changes:*
- Add job to notify the PAD paid events to partner
- Change reconcilation to suppress express checkout PAD events
- Add partner notified at to the payment link


pay-api/src/pay_api/services/pad_service.py
```diff
@@ -154,6 +155,15 @@ def complete_post_invoice(
         **kwargs,  # noqa: ARG002
     ) -> None:
         """Complete any post invoice activities if needed."""
+        # Express-checkout PAD: hold the create-time notification. Customer's money hasn't moved yet
+        # (CFS ack + reversal window still pending); the drain job publishes once safe.
+        # Reversals go directly through release_payment_or_reversal from refund.py and are unaffected.
+        if CodeService.is_express_checkout_enabled(invoice.corp_type_code):
```
Jxio: If a corp type has express checkout enabled, any regular PAD invoice for that corp type (no link row) has its create-time publish suppressed but can never be picked up by the drain job?
Should gate on "does this invoice have a link" instead of the corp-type flag?

jobs/payment-jobs/tasks/express_checkout_pad_notify_task.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+
+        published = 0
+        for invoice in candidates:
+            link = (
+                db.session.query(InvoicePaymentLinkModel)
+                .filter(InvoicePaymentLinkModel.invoice_id == invoice.id)
+                .filter(InvoicePaymentLinkModel.partner_notified_at.is_(None))
+                .first()
+            )
```
Jxio: why not move this to candidates?
```
candidates = (
    InvoiceModel.query
    .join(InvoicePaymentLinkModel, InvoicePaymentLinkModel.invoice_id == InvoiceModel.id)
    .filter(status == PAID, ...)
    .filter(InvoicePaymentLinkModel.partner_notified_at.is_(None))
    .all()
)
```

pay-api/src/pay_api/services/pad_service.py
sumesh85: My thought process was these corp types would only use the express checkout, coz the reason why they are using express checkout is because they cannot directly integrate with our keycloak and need to create it externally. However, it's good to implement it in a way that would work in either scenarios. Will update the PR.

jobs/payment-jobs/tasks/express_checkout_pad_notify_task.py
sumesh85: Updated

pay-api/src/pay_api/services/pad_service.py
sumesh85: Updated

Jxio (approved):
lgtm, approve, just need fix lint and unit test

ochiu (approved):

PR #2612, opened 2026-08-12
fix(pay-api): replace wildcard CORS with configurable origin allow-list

panish16:
## Summary

Pay-API's CORS configuration allowed **any** origin:

- The global `after_request` hook in `pay_api/__init__.py` unconditionally set `Access-Control-Allow-Origin: *` on every response (and again in the 403 handler).
- Every route across 20 resource files used `@cross_origin(origins="*", ...)`.
- The existing test suite explicitly asserted `*` as correct behaviour.

Since Keycloak JWTs are bearer tokens attached by client-side JS, a wildcard CORS policy means any external site's JavaScript could call Pay-API with a user's token and read the response — Keycloak's redirect/web-origin restrictions only govern where a token can be *issued*, not where it can be *used*.

This PR replaces the wildcard with a configurable allow-list:

- Added `CORS_ORIGINS` app setting (`pay_api/config.py`), populated from a comma-separated `CORS_ORIGINS` env var. Empty by default → fails closed (blocks all cross-origin browser requests) rather than falling back to permissive behaviour.
- `pay_api/__init__.py`: the global header hook and the 403 handler now only echo back `Access-Control-Allow-Origin` (with `Vary: Origin`) when the request's `Origin` matches the configured allow-list.
- Removed `origins="*"` from all 20 `@cross_origin(...)` decorators so they inherit the app-level `CORS_ORIGINS` config (flask-cors reads `app.config["CORS_ORIGINS"]` automatically when `origins` isn't passed explicitly).
- Added `CORS_ORIGINS` to `devops/vaults.gcp.env` and `devops/vaults.ocp.env` under the existing `relationship/$APP_ENV/pay-api/` 1Password path, matching this repo's existing per-env secret pattern.
- Updated `tests/unit/api/test_cors_preflight.py` to assert against a configured test origin instead of `*`, and added two new tests asserting an allowed origin is echoed back (with `Vary: Origin`) and a disallowed origin gets no `Access-Control-Allow-Origin` header at all.

## Domain audit — how the allow-list was built

Pay-API's browser-facing consumers turned out to be much broader than expected. Started from Keycloak client configs and pay-api's own `VALID_REDIRECT_URLS`, then verified every candidate by reading the **actual runtime config embedded in the live JS/HTML served to real browsers** — `.env.example` files and even some Keycloak `webOrigins` entries turned out to be stale or misleading, so nothing below is based on assumption alone.

**12 distinct frontends confirmed as genuine direct (non-gateway) callers of Pay-API**, live-verified per environment:

| App | Origin pattern |
|---|---|
| `pay-ui` (fas-web — Pay-API's own routing-slip/fee-accounting frontend) | `{env}.pay.bcregistry.gov.bc.ca`, `fas{,2}-{env}.apps.silver.devops.gov.bc.ca`, `fas-{env}.web.app` |
| `ppr-ui` | `{env}.assets.bcregistry.gov.bc.ca` |
| `registries-search` | `{env}.search.bcregistry.gov.bc.ca` |
| `bcregistry` (homepage) | `{env}.bcregistry.gov.bc.ca` |
| `documents-ui` | `documents-ui-{env}.web.app`, `{env}.documents.bcregistry.gov.bc.ca` |
| `btr-web` | `bcregistry-btr-{env}.web.app` |
| `bor-ui` | `bcregistry-bor-{env}.web.app` |
| STRR (5 sub-apps: main, examiner, strata, platform, host-pm) | `strr-{app}-{env}.web.app`, `{env}.{platform,stratahotel,host}.shorttermrental.registry.gov.bc.ca` |

Plus, confirmed via pay-api's own **live** `VALID_REDIRECT_URLS` (dev/test only — not present in prod's):
- `www.bcregistry.daxiom.ca` (Daxiom — bcgov contractor, document-delivery integration for MHR/PPR/CSO)
- `assessmentreportsonline-{dev,test}.bcassessment.ca` (BC Assessment's paid report-ordering portal)

And the `bcregistry.ca` domain family (`{env}.bcregistry.ca` / `www.bcregistry.ca`), confirmed via the same live `VALID_REDIRECT_URLS` evidence.

**Confirmed explicitly NOT needed** (verified in their actual served bundles, not assumed):
- `business-create-ui`, `business-edit-ui`, `auth-web`, `business-ui` — all call Pay-API through the API gateway (`api.connect.gov.bc.ca/pay-{env}`), confirmed baked into their real JS bundles.
- `pay-admin` — imports `pay_api.models` and talks to Pay-API's database directly, server-side; never calls the HTTP API from a browser.
- Various backend/job/queue services (lear, namex, sbc-auth, ppr's/registries-search's/STRR's own APIs and queues) — server-to-server, no browser `Origin` involved even though several use the same direct pay-api URL.

**Sandbox note:** most of the 12 apps aren't deployed to sandbox yet (confirmed — their Keycloak-configured sandbox origins resolve in DNS but return a genuine Firebase 404, not a connection failure). Sandbox's `CORS_ORIGINS` includes the full Keycloak-documented intended set anyway (zero risk to include an unused origin; the alternative is a silent break the moment one of these apps is deployed there).

## `CORS_ORIGINS` values — set in 1Password (relationship/{env}/pay-api/CORS_ORIGINS)

Populated for dev/test/sandbox/prod as of this PR. Not reproduced here since they'll drift from this description over time — 1Password is the source of truth going forward, not this PR body.

## Before merging / deploying

- `CORS_ORIGINS` must be set in the vault for an environment **before** it's deployed with this change — empty/missing fails closed and immediately breaks every frontend listed above in that environment.
- Recommend rolling out dev → test → sandbox → prod, confirming after each: known frontends still work, and a spoofed `Origin` header gets no `Access-Control-Allow-Origin` back.

## Separate finding, not part of this PR

While verifying gateway-routed frontends didn't need entries here, tested the API gateway's (`api.connect.gov.bc.ca`) own CORS handling directly — it reflects back **any** `Origin` unconditionally, with `Authorization` in the allowed headers list. Same class of vulnerability as this PR fixes, one layer up, and this PR does nothing to protect gateway-routed traffic. Flagging separately to the gateway owner; not fixed here since it's out of scope for pay-api.

## Test plan

- [x] `python -m py_compile` on all changed files
- [x] `ruff format --check .` / `ruff check .` — full repo, clean
- [x] `tests/unit/api/test_cors_preflight.py` — 20/20 passing (including 2 new tests) against a real running app + real Postgres, not mocked
- [x] Full `tests/unit` suite run for regressions — confirmed identical failure set to baseline `main` (pre-existing infra/mock-stack gaps, unrelated to this change), 2 more passing (the new CORS tests)
- [ ] Manual regression per environment after deploy (dev → test → sandbox → prod), per "Before merging / deploying" above

Relates to bcgov/entity#32488 (SRE audit ticket).


sumesh85 (approved):
Approving the PR to test in dev

PR #2626, opened 2026-08-27
34723: Support stamping payment method to invoice without changing the account

sumesh85:
*Issue #:*
https://github.com/bcgov/entity/issues/34723

*Description of changes:*


pay-api/src/pay_api/resources/v1/account.py
```diff
@@ -138,14 +138,16 @@ def put_account(account_number: str):
     """Update the payment account records."""
     current_app.logger.info(f"<put_account {account_number}")
     request_json = request.get_json()
-    current_app.logger.debug(request_json)
+    scope_value = request.args.get("scope", None)
```
sumesh85, line 141: Added scope param to decide whether the change needs to be applied everywhere. Now it supports cfs_account, we could use comma separated items to enhance this

pay-api/src/pay_api/services/online_banking_service.py
```diff
@@ -43,16 +43,42 @@ def get_payment_method_code(self):
...
         cfs_account = CfsAccountModel()
-        cfs_account.status = CfsAccountStatus.PENDING.value
         cfs_account.payment_method = PaymentMethod.ONLINE_BANKING.value
+
+        if not contact_info:
+            cfs_account.status = CfsAccountStatus.PENDING.value
+            return cfs_account
+
+        try:
+            details = self.create_cfs_account(
```
sumesh85, line 65: We will force creation of an account here so that users can get the CFS number for payment, there would be cases where CFS is down (scheduled) where we cannot provide this. Will need to ask the use to come back later on the web.

ochiu (approved):

Jxio (approved):

PR #2628, opened 2026-08-31
34417 - account link auth check for invoice / receipt

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/34417

- refactor account-linking-key logic to be opt-in for check auth to limit usage to specific operations

### DIFF
--- a/pay-api/src/pay_api/services/invoice.py
+++ b/pay-api/src/pay_api/services/invoice.py
@@ -408,7 +408,12 @@ class Invoice:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         return invoice
 
     @staticmethod
-    def find_by_id(identifier: int, skip_auth_check: bool = False, one_of_roles=ALL_ALLOWED_ROLES):
+    def find_by_id(
+        identifier: int,
+        skip_auth_check: bool = False,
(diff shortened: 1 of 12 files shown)


pay-api/src/pay_api/services/invoice.py
```diff
@@ -596,8 +601,12 @@ def create_invoice_pdf(identifier: int, **kwargs) -> tuple:
 
     @staticmethod
     @user_context
-    def _check_for_auth(dao, one_of_roles=ALL_ALLOWED_ROLES, **kwargs):
+    def _check_for_auth(dao, one_of_roles=ALL_ALLOWED_ROLES, allow_linking_key: bool = False, **kwargs):
         user: UserContext = kwargs["user"]
+        # Check auth by business identifier if account linking key is also present.
+        if allow_linking_key and user.linking_key and dao.business_identifier:
+            check_auth(dao.business_identifier, one_of_roles=one_of_roles, allow_linking_key=True)
```
ochiu, line 608: Different operations use this related to invoices. Added allow_linking_key on check_for_auth and check_auth with a default of False (exclude linking key header default) to allow for opt-in usage of the linking key.

sumesh85 (approved):

PR #2644, opened 2026-09-14
34853 - Keep the adhoc service account off payer-facing

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/34853

*Description of changes:*


pay-api/src/pay_api/utils/util.py
```diff
@@ -435,3 +437,19 @@ def get_statement_currency_string(value):
...
+    """Blank the adhoc SA account on a serialized invoice, in place.
+
+    Express-checkout invoices are parked on `sa-<client_id>` until a payer redeems the
+    link, and an anonymous payer never redeems — so the account stays on that adhoc row
+    forever. It is internal routing, meaningless to a payer, and it surfaces as
+    "Account Number: sa-..." on receipts. A redeemed invoice carries a real account and
+    is left untouched.
+    """
+    account = invoice.get("payment_account") or {}
+    if str(account.get("account_id") or "").startswith(EXPRESS_CHECKOUT_ACCOUNT_PREFIX):
```
sumesh85: I think checking this against the prefix is not sustainable solution, if the prefix changes for some reason in future this check would fail for old receipts.
Here the account is a fake account and doesn't exist in PaymentAccount. So a better solution would be to check if the account exists in the table, than using a prefix check.

Jxio: just check the payment_accounts table, it is exist. 
I think it generated from: https://github.com/bcgov/sbc-pay/blob/cd27de5523f746971309e59f499764e9f649dc14/pay-api/src/pay_api/services/payment_service.py#L223

Jxio: how about check if linked_at is None?

sumesh85: Yes, right the account gets created. Or simply we can check if it is not a number (as all real accounts would have the auth_account_id as a number from auth) ?

Jxio: hmm I found some non-numeric auth_account_id on db:
```
select auth_account_id,
       count(*) as accounts
from payment_accounts
where auth_account_id !~ '^[0-9]+$'
group by auth_account_id
order by accounts desc;
```

Jxio: @sumesh85 updated, that's a good catch!

pay-api/src/pay_api/models/invoice_payment_link.py
```diff
@@ -54,3 +54,9 @@ class InvoicePaymentLink(BaseModel):
     def find_by_token(cls, token: str) -> "InvoicePaymentLink | None":
         """Return the link row for a token, regardless of state."""
         return cls.query.filter_by(token=token).one_or_none()
+
+    @classmethod
+    def is_unredeemed_for_invoice(cls, invoice_id: int) -> bool:
+        """Return True if the invoice has a link row nobody has claimed yet."""
+        row = cls.query.with_entities(cls.linked_at).filter_by(invoice_id=invoice_id).first()
```
sumesh85: Can we replace first() with one_or_none() as we expect atmost 1 record for this recod

sumesh85 (approved):

pay-api/src/pay_api/models/invoice_payment_link.py
Jxio: yup!

PR #2647, opened 2026-09-15
34847: Changes to add email and returnUrl to the express pay invoices

sumesh85:
*Issue #:*
https://github.com/bcgov/entity/issues/34847

*Description of changes:*
- Adding email and returnUrl to the express checkout ivnoices
- Some test cleanup


pay-api/src/pay_api/services/payment_link.py
```diff
@@ -152,7 +157,10 @@ def find_invoice_by_token(cls, token: str) -> dict:
         link = cls.resolve_token(token, allow_linked=True)
         skip_auth = link.linked_at is None
         invoice = InvoiceService.find_by_id(link.invoice_id, skip_auth_check=skip_auth)
-        return invoice.asdict(include_dynamic_fields=True)
+        result = invoice.asdict(include_dynamic_fields=True)
+        if link.return_url:
+            result["returnUrl"] = link.return_url
```
Jxio, line 162: is `returnUrl` suppose run through the same validation as `clientSystemUrl`?

sumesh85: Yes we could add it. For the client system url, this is something we use in payment transactions POST endpoint which is a public endpoint and that's where we need the allowed validations to avoid phishing. 
For payment link, this happens with a curated service account, But yea, adding this won't hurt.

sumesh85: Updated

Jxio (approved):

ochiu (approved):

PR #2648, opened 2026-09-15
34532 - Receipt notifications to the account owner/coordinator

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/34532

*Description of changes:*


pay-api/src/pay_api/services/email_service.py
```diff
@@ -146,6 +149,78 @@ def _render_credit_add_notification_template(params: dict) -> str:
...
+    OB and PAD later in reconciliation.
+
+    Skipped for unredeemed express-checkout invoices: there is no real account to notify,
+    and the payer reaches their receipt through the link. Mail failures are logged and
+    never affect the payment.
+    """
+    try:
+        payment_account = invoice.payment_account
+        auth_account_id = payment_account.auth_account_id if payment_account else None
+        if not auth_account_id or InvoicePaymentLinkModel.is_unredeemed_for_invoice(invoice.id):
```
sumesh85: InvoicePaymentLink contains email now, so even if there is no account we would need to send if email is present

Jxio: added

sumesh85 (approved):

pay-api/src/pay_api/services/auth.py
```diff
@@ -153,14 +153,18 @@ def check_auth(
 
 
 @user_context
-def get_account_admin_users(auth_account_id, use_service_account=False, **kwargs):
-    """Retrieve account admin users."""
+def get_account_admin_users(auth_account_id, use_service_account=False, roles: str = "ADMIN", **kwargs):
```
ochiu: hmmm this isn't really just get account_admin_users any more. Do we use this in other components? easy to rename?

ochiu: If there are a lot of dependencies, for a clean transition we can keep this function, but create a get_account_users_by_role and move the logic to there. Then get_account_admin_users would just call the new function feeding in only ADMIN

Jxio: there're other 12 places using it, I'm thinking add a new get_account_members might be better

ochiu: sounds good

ochiu (approved):

